PR #1, opened 2017-09-14
Add Campus Authentication

srkirkland:


jcubbage:
Need to set IP port for devel server under project settings to 58369.

PR #4, opened 2017-09-19
Jpk reactify

jpknoll:


jcubbage:
When I run this and go to http://localhost:58369/home/react, I hit an error in vendor.js, line 31596, could not find module "react-router-dom". When I goggled, looksed like it was a dependency that react-router forgot to put in their dependencies, so npm doesn't download. Ticket I read said it was fixed, but maybe not in the version we're using. Should react-router-dom be added to the package.json? Or maybe update react-router?

srkirkland:
@jcubbage Well we aren't using react-router anyway so maybe @jpknoll can fix that

Keas.Mvc/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "sloth-web",
```
jSylvestre: @jpknoll Sloth/Keas...

jpknoll:
@jcubbage Give it another go. I need to remove one last reference in the webpack configuration.

jcubbage:
Now getting JS error at line 79, column 154 in main.js. Just says syntax error. That line is the Hot Module Replacement line, so possible something else isn't installed yet? Not sure it if this is relevant, but I'm running without local admin. Had to temporarily grant admin to get node.js installed and npm happy.

jcubbage:
Windows 8 sucks apparently.

PR #6, opened 2017-09-19
Entity framework setup

srkirkland:


Keas.Mvc/Startup.cs
```diff
@@ -24,6 +26,9 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+            // setup entity framework
+            services.AddDbContextPool<ApplicationDbContext>(o => o.UseSqlite("Data Source=keas.db"));
```
jSylvestre: @srkirkland Path it so it doesn't end up in the bin directory? And add to gitignore...

srkirkland: @jSylvestre it is in the gitignore now, and at least in my build it ends up in the web root and not the bin directory, which is a decent place i think.

jcubbage:
Did test, and looks good.

PR #8, opened 2017-09-22
Auth claims and IAM

srkirkland:
Note, you should add an “IamKey” to your app settings so you can use IAM


jcubbage:
On IdentityService, when I but a break on line 35, I see a value in results.responseData.results, but the code is finding it to be null and returning string.Empty. If I comment out this line, UserId on Line72 of startup has the correct IamID for me. ?????

jcubbage:
Working great now!

PR #9, opened 2017-09-29
Jsc/poco

jcubbage:
This looks like a good first pass out our initial models. Added initializers for all the tables.

### DIFF
--- a/Keas.Core/Domain/Location.cs
+++ b/Keas.Core/Domain/Location.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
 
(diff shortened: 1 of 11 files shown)


Keas.Core/Domain/Location.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Keas.Core.Domain
+{
+    public class Location
```
srkirkland: @jcubbage i forget is a location supposed to be tied to a team or are they system-wide

jSylvestre: I think we said they were tied to a team...

jcubbage:
Updated key references as requested. Inverse property not working on TeamMembers to Assignments and Assets. Not sure what I'm messing up, but I'm done for today. :)

srkirkland:
@jcubbage We'll have to tweak all this stuff anyway so for now I say 🚢

PR #15, opened 2017-11-30
fetches and other fixes

srkirkland:


jcubbage:
Can we keep the random key assignment for all assets? Just go in and randomly assign random things to random people?

PR #28, opened 2018-01-25
Lh/equipment

laholstege:
add basic equipment
add action dropdown


Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    this.state = {
+        equipment: [],
+      loading: true
+    };
+  }
+  public async componentDidMount() {
+    // are we getting the person's equipment or the team's?
+    const equipmentFetchUrl = this.context.person
+      ? `/equipment/listassigned/${this.context.person.id}`
```
srkirkland: @laholstege we've been forgetting but remember we'll have to put the team name in the URL so the controller knows what team we are getting the equipment for (not strictly needed for listassigned but needed for list, and good for both.

Keas.Mvc/Controllers/EquipmentController.cs
```diff
@@ -0,0 +1,81 @@
...
+    public class EquipmentController : SuperController
+    {
+        private readonly ApplicationDbContext _context;
+
+        public EquipmentController(ApplicationDbContext context)
+        {
+            this._context = context;
+        }
+
+        public string GetTeam()
```
srkirkland: @laholstege let's dump this in the future, it doesn't do anything useful

PR #29, opened 2018-02-28
Lh/equipment updates

laholstege:
search for equipment to add
can also revoke/add/show details for item in list with action button

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
@@ -19,8 +19,9 @@ import 'react-datepicker/dist/react-datepicker.css';
 
 
 interface IProps {
-    onAssign: (equipment: IEquipment, date: string) => Promise<IEquipmentAssignment>;
-    assignedEquipmentList: string[];
+    onCreate: (equipment: IEquipment, person: IPerson) => Promise<IEquipment>;
+    assignedEquipmentList?: string[];
+    allEquipmentList?: IEquipment[];
 }
 
 interface IState {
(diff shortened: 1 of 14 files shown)


Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
```diff
@@ -1,87 +1,184 @@
...
-    // TODO: check for success
-    this.setState({ modal: false });
+  private _onSelectPerson = (person: IPerson) => {
+      this.setState({ person }, this._validateState);
   };
+
+
+  private _validateState = () => {
+      let valid = true;
+      if (this.props.selectedEquipment == null)
```
srkirkland: @laholstege for these, try to use truthy values, like `if(!this.props.selectedEquipment)`.  That way it'll catch things like undefined.  Also, try to always use === or !== when comparing, so we don't get javascript trying to do some explicit conversions (like with the error message)

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -38,63 +51,141 @@ export default class EquipmentContainer extends React.Component<{}, IState> {
...
       });
-    }
+      }
+
+    let index = this.state.equipment.findIndex(x => x.id == equipment.id);
+    console.log("index " + index);
+    if (index != -1) {
+        console.log("changing");
+        //update already existing entry in equipment
+        let updateEquipment = this.state.equipment.slice();
```
srkirkland: @laholstege interesting, is this your way of making a shallow copy?  I've never seen anything like this but i guess it's the same as `let updateEquipment = [...this.state.equipment]`.  Is that what is intended?

Keas.Mvc/Helpers/DbInitializer.cs
```diff
@@ -15,7 +15,7 @@ public static void Initialize(ApplicationDbContext context)
             if (context.Users.Any()) return; // already initialzied
 
             var scott = new User { Id = "123124", FirstName = "Scott", Name = "Scott Kirkland", Email = "scott@email.com" };
-            var jason = new User { Id = "123222", Name = "Jason", Email = "jsylvestre@email.com" };
+            var jason = new User { Id = "123222", Name = "Jason", Email = "jason@email.com" };
```
srkirkland: @laholstege ha i like this one

PR #31, opened 2018-03-13
Lh/equipment

laholstege:
fixes searching
adds date on assign
adds sn/make/model on create

still have to decide how we want to allow editing after creation, but right now details and edit are the same component so it should be easy to implement

the styling is way better now but still need to get the datepicker input to be less ugly


Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
```diff
@@ -149,18 +163,24 @@ export default class AssignEquipment extends React.Component<IProps, IState> {
...
                     </div>
-
-
-                    {this.props.selectedEquipment != null &&
+                    {!this.props.selectedEquipment || !this.props.selectedEquipment.teamId && //if we are creating a new equipment, edit properties
+                        <EquipmentEditValues selectedEquipment={this.props.selectedEquipment} changeProperty={this.props.changeProperty} disableEditing={false} />
+                    }
+                    {this.props.selectedEquipment && !!this.props.selectedEquipment.teamId &&
+                         <EquipmentEditValues selectedEquipment={this.props.selectedEquipment} disableEditing={true} />
+                    }
```
srkirkland: @laholstege this works pretty well, I think it would be cleaner though if just in the render method you added a bit of code to determine outside of the HTML portion if we are editing or not.  Just like `const editable = <condition>` then you can reuse that and probably make the render easier too

PR #34, opened 2018-03-22
bring in facilities link info

srkirkland:
names aren't great but it should work for read-only querying


jcubbage:
Looks good for first pass. Hopefully we can keep this one-directions and simple.

PR #35, opened 2018-03-27
Introduce router to asset page

srkirkland:
Now we get full route support for main asset tabs as well as add/details modals.


srkirkland:
Well this changes a lot of things, and breaks stuff too.  One big difference is that person is not part of this.context any more, since we are going to be dynamically loading people (not just one on its own page).  We'll have to go through and pass person as part of props where appropriate, and update the route methods to match

PR #44, opened 2018-04-10
Lh/assign room

laholstege:
adds actually room functionality to equipment. is not related to space as of right now.
keys has an assign room that doesn't actually do anything since we will need to change it to have many rooms


Keas.Core/Domain/Key.cs
```diff
@@ -9,6 +9,7 @@ public class Key : AssetBase {
         public string SerialNumber { get; set; }
 
         public string LockId { get; set; }
+        public Room Room { get; set; }
```
srkirkland: @laholstege this won't be the case that a key has one room associated, maybe not even any via this mechanism.

PR #45, opened 2018-04-12
Jsc/team

jcubbage:
I think this completes all things Team/TeamAdmin on my list. Next stop....securityservice to check if user is in correct role for action being taken. Yeah!

On a side note, Bootstrap dropped glyphicons. Do we want to bring in Font Awesome, glyphicons, or GitHub Octicons (https://octicons.github.com/)?


jcubbage:
Also.....I think we should merge and delete, then create new branch for the security service. Thoughts?

srkirkland:
@jcubbage we are already using font awesome, so just use that one

PR #47, opened 2018-04-19
Lh/spaces

laholstege:
details modal pulls equipment and keys assigned to the room. 
currently does not check teams, just pulls based on roomKey

still need to think about showing # of keys/equip on table with fancy linq


Keas.Mvc/Keas.Mvc.csproj
```diff
@@ -5,6 +5,10 @@
    <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
     <TypeScriptToolsVersion>Latest</TypeScriptToolsVersion>  
   </PropertyGroup>
+  <ItemGroup>
+    <None Remove="ClientApp\components\Spaces\SpacesContainer.tsx" />
+    <None Remove="ClientApp\components\Spaces\SpacesListItem.tsx" />
+  </ItemGroup>
```
srkirkland: @laholstege do these really need to be in the csproj?  Same with the later ones that are set to typescript compile.  why do we need them and just these?

laholstege: @srkirkland no vs added it and i forgot to remove it

PR #50, opened 2018-04-25
Jsc/history2

jcubbage:
good spot to bring this in. Added credentials to fetch api & [authorize] to SuperController.

### DIFF
--- a/Keas.Core/Domain/History.cs
+++ b/Keas.Core/Domain/History.cs
@@ -20,6 +20,12 @@ namespace Keas.Core.Domain {
         public Person Person { get; set; }
         public int PersonId { get; set; }
 
+        // Key vs Equipment vs Access 
+        public string AssetType { get; set; }
+
+        // Created, granted, accepted, declined, revoked, deleted 
+        public String ActionType { get; set; }
+
         public Key Key {get;set;}
         public Equipment Equipment { get; set; }
 
--- a/Keas.Mvc/Controllers/KeysController.cs
+++ b/Keas.Mvc/Controllers/KeysController.cs
@@ -68,7 +68,9 @@ namespace Keas.Mvc.Controllers
             {
                 _context.Keys.Add(key);
                 await _context.SaveChangesAsync();
-                var userName = this.User.Identity.Name;
+
+                //none of these has values?????
+                var userName = CurrentUserId;
                 var userid = _securityService.GetUser();
                 var user = await _context.Users.SingleOrDefaultAsync(x => x.Email == this.User.FindFirstValue(ClaimTypes.NameIdentifier)); ;
                 await _eventService.TrackCreateKey(key, user);
--- a/Keas.Mvc/Controllers/KeysController.cs
+++ b/Keas.Mvc/Controllers/KeysController.cs
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Security.Claims;
 using System.Threading.Tasks;
 using Keas.Core.Data;
 using Keas.Core.Domain;
+using Keas.Mvc.Services;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
@@ -14,10 +16,14 @@ namespace Keas.Mvc.Controllers
     public class KeysController : SuperController
     {
         private readonly ApplicationDbContext _context;
+        private readonly IEventService _eventService;
+        private readonly ISecurityService _securityService;
 
-        public KeysController(ApplicationDbContext context)
+        public KeysController(ApplicationDbContext context, IEventService eventService, ISecurityService securityService)
         {
             this._context = context;
+            _eventService = eventService;
+            _securityService = securityService;
         }
 
         public string GetTeam()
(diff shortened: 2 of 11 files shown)


Keas.Mvc/Controllers/SuperController.cs
```diff
@@ -1,17 +1,18 @@
...
 namespace Keas.Mvc.Controllers
 {
+    [Authorize]
     public class SuperController : Controller
     {
         public string Team => ControllerContext.RouteData.Values["teamName"] as string;
 
         private const string TempDataMessageKey = "Message";
 
+        public string CurrentUserId => User.FindFirst(ClaimTypes.NameIdentifier)?.Value;
```
jSylvestre: @jcubbage Did you test this?

Keas.Core/Domain/History.cs
```diff
@@ -20,6 +20,12 @@ public History()
         public Person Person { get; set; }
         public int PersonId { get; set; }
 
+        // Key vs Equipment vs Access 
+        public string AssetType { get; set; }
+
+        // Created, granted, accepted, declined, revoked, deleted 
+        public String ActionType { get; set; }
```
srkirkland: @jcubbage lowercase type please

Keas.Mvc/Services/SecurityService.cs
```diff
@@ -108,7 +113,29 @@ private bool IsInRole(Role role, Team team)
             var user = await _dbContext.Users.SingleOrDefaultAsync(x => x.Email == userId);
             return user;
         }
-        
+
+        public async Task<List<User>> GetUsersInRoles(List<Role> roles, int teamId)
```
srkirkland: @jcubbage this doesn't seem to do what it says -- I don't see where you are finding the users in the given list of roles.  Maybe you need a .contains in the where statement?

Keas.Mvc/Controllers/SuperController.cs
jcubbage: I did....it's not very useful. I think it give you the IAM ID. I forgot to back it out. I've submitted another commit to reverse. The trick is Scott's login sequence takes the email address passed from authentication and stores it as claim. I use it to get user in the security service. Wasn't working because the API wasn't passing credentials.

srkirkland: @jcubbage Note that (1) I store a bunch of claims, so other ones might be helpful, but more importantly (2) we'll probably change auth to go back to CAS which means the claims will change a bit anyway.  you are probably better off, if you need something like this, to use the UserManager class to get the current user

Keas.Mvc/Services/SecurityService.cs
jcubbage: Dropped that when I refactored it. Was originally:
using (_dbContext) 
            { 
                var team = await _dbContext.Teams.SingleAsync(t => t.Id == teamId); 
 
                _dbContext.Entry(team) 
                    .Collection(t => t.TeamPermissions) 
                    .Query() 
                    .Where(tp => roles.Contains(tp.Role)) 
                    .Include(tp => tp.User) 
                    .Load(); 
....}

Keas.Mvc/Services/NotificationService.cs
```diff
@@ -68,7 +68,7 @@ public NotificationService(ApplicationDbContext dbContext, ISecurityService secu
             var users = await _securityService.GetUsersInRoles(roles, access.TeamId);
             foreach (var user in users)
             {
-                var emailQueue = new EmailQueue
+                var emailQueue = new Notification
```
srkirkland: @jcubbage change the var name too

jcubbage: Lots of changes!

Keas.Mvc/Controllers/KeysController.cs
```diff
@@ -62,6 +68,7 @@ public string GetTeam()
             {
                 _context.Keys.Add(key);
                 await _context.SaveChangesAsync();
+                await _eventService.TrackCreateKey(key, _securityService.GetUser().Result);
```
srkirkland: @jcubbage here you should use await as well. never .Result, it's unreliable in an MVC environment.

`await _eventService.TrackCreateKey(key, await _securityService.GetUser());`

Keas.Core/Domain/History.cs
```diff
@@ -7,23 +7,34 @@ public History()
...
             ActedDate = DateTime.UtcNow;
         }
+        [Key]
         public int Id { get; set; }
 
         [Required]
         public string Description { get; set; }
 
         public User Actor { get; set; }
-        public string ActorName { get; set; }
```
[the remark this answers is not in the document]
srkirkland: @jcubbage actually, as long as we put the name in the description I guess we can leave it out of here for now

Keas.Mvc/Controllers/TestSecurityController.cs
```diff
@@ -20,12 +20,11 @@ public TestSecurityController(ISecurityService securityService, ApplicationDbCon
             _securityService = securityService;
             _context = context;
         }
-        public IActionResult Index()
+        public async Task<IActionResult> Index()
         {
             var roles = _context.Roles
                 .Where(r => r.Name == Role.Codes.KeyMaster || r.Name == Role.Codes.DepartmentalAdmin).ToList();
```
srkirkland: @jcubbage in the future make all of these ToListAsync() and use await

Keas.Core/Domain/History.cs
```diff
@@ -7,23 +7,34 @@ public History()
         {
             ActedDate = DateTime.UtcNow;
         }
+        [Key]
```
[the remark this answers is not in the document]
jcubbage: Got it. Was struggling to get this table to work (would make object, but not save it). Turned out to be all the int? on the assets.

PR #51, opened 2018-04-25
Pushed stylesheet to CDN

klanmiko:


klanmiko:
@srkirkland oh yea, that's totally possible, I would just need to move the media folder and it would include the SVG background and fonts

srkirkland:
@klanmiko cool yea i think that's a good plan

PR #52, opened 2018-05-02
Jsc/history2

jcubbage:
Tracks all currently available actions. Needs to expand when we get update and delete working for each asset as well as people accepting assignments.

### DIFF
--- a/Keas.Mvc/Controllers/AccessController.cs
+++ b/Keas.Mvc/Controllers/AccessController.cs
@@ -125,6 +125,7 @@ namespace Keas.Mvc.Controllers
             {
                 _context.AccessAssignments.Remove(accessAssignment);
                 await _context.SaveChangesAsync();
+                await _eventService.TrackUnAssignAccess(accessAssignment, await _securityService.GetUser());
                 return Json(accessAssignment);
             }
             return BadRequest(ModelState);
--- a/Keas.Mvc/Controllers/AccessController.cs
+++ b/Keas.Mvc/Controllers/AccessController.cs
@@ -109,8 +109,10 @@ namespace Keas.Mvc.Controllers
                     PersonId = personId,
                     ExpiresAt = DateTime.Parse(date),
                 };
+                accessAssingment.Person = await _context.People.Include(p => p.User).SingleAsync(p => p.Id == personId);
                 _context.AccessAssignments.Add(accessAssingment);
                 await _context.SaveChangesAsync();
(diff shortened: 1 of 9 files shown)


Keas.Mvc/Controllers/AccessController.cs
```diff
@@ -88,8 +93,10 @@ public string GetTeam()
             {
                 _context.Access.Add(access);
                 await _context.SaveChangesAsync();
+                await _eventService.TrackCreateAccess(access, await _securityService.GetUser());
+                return Json(access);
             }
-            return Json(access);
```
jcubbage: @laholstege Can you confirm this change is correct?

laholstege: @jcubbage looks good

Keas.Mvc/Controllers/EquipmentController.cs
```diff
@@ -98,6 +109,7 @@ public string GetTeam()
                 _context.EquipmentAssignments.Remove(eq.Assignment);
                 eq.Assignment = null;
                 await _context.SaveChangesAsync();
+                await _eventService.TrackUnAssignEquipment(equipment, await _securityService.GetUser());
```
srkirkland: @jcubbage why not have the event service call the security service itself?  Seems like it would save a lot of typing and extra service instantiations

Keas.Mvc/Services/EventService.cs
```diff
@@ -24,6 +32,57 @@ public EventService(IHistoryService historyService, INotificationService notific
...
+        }
+
+        public async Task TrackAssignAccess(AccessAssignment accessAssignment, User user)
+        {
+            var history = await _historyService.AccessAssigned(accessAssignment, user);
+            await _notificationService.AccessAssigned(accessAssignment, history);
+
+        }
+
+        public async Task TrackUnAssignAccess(AccessAssignment accessAssignment, User user)
```
srkirkland: @jcubbage since the initiator is always the currently logged in user, shouldn't we just grab via the security service here?  Or optionally maybe just grab if passed in user == null?

Keas.Mvc/Controllers/EquipmentController.cs
srkirkland: @jcubbage cool, i'd say give it a quick try and see if it's working now that credentials are being passed, and if no don't worry about it

PR #53, opened 2018-05-03
Lh/attributes

laholstege:


Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
```diff
@@ -35,8 +35,8 @@ export default class AccessContainer extends React.Component<IProps, IState> {
   public async componentDidMount() {
       // are we getting the person's access or the team's?
       const accessFetchUrl = this.props.person
-          ? `/access/listassigned?personId=${this.props.person.id}&teamId=${this.props.person.teamId}`
-      : `/access/list?teamId=${this.context.team.id}`;
+          ? `/api/${this.context.team.name}/access/listassigned?personId=${this.props.person.id}&teamId=${this.props.person.teamId}`
+      : `/api/${this.context.team.name}/access/list?teamId=${this.context.team.id}`;
```
jcubbage: Is this redundant? I think it might be better as `/api/${this.context.team.name}/access/list? and update the action to look for x.team.Name = Team instead of x.Team.Id. Same with listassigned passing both team.name and teamId.

srkirkland: true

Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
```diff
@@ -165,9 +187,17 @@ export default class AssignEquipment extends React.Component<IProps, IState> {
...
     }
 
+    console.log("in assign");
+    console.log(this.state.equipment);
+
     const person = this.props.person ? this.props.person : this.state.person;
 
-    await this.props.onCreate(person, this.state.equipment, this.state.date.format());
+    const equipment = this.state.equipment;
+    equipment.attributes = equipment.attributes.filter(x => !!x.key);
```
srkirkland: yo quiero

Keas.Mvc/ClientApp/components/Equipment/EquipmentAttributes.tsx
```diff
@@ -0,0 +1,85 @@
...
+    this.props.updateAttributes(attributes);
+  }
+
+  private _onEditAttribute = (i: number, prop: string, val: string) => {
+    const attributes = this.props.equipment.attributes;
+    attributes[i] = {
+      ...attributes[i],
+      [prop]: val
+    }
+    this.props.updateAttributes(attributes);
```
srkirkland: @laholstege I think this should work but in react sometimes it won't notice the change because your attributes will still be the same list.  Same w/ splice for removing.  If it works that's fine, but your changes lists will be reference equals so like if you are using redux for example it won't notice the prop has changed and should be updated

PR #54, opened 2018-05-04
Confirm

jcubbage:
That seemed too easy. I'm sure I'm missing something. Might be nice to have @cydoval  make this look better than I came up with. Or will he just go through later and clean up my lame UI?

### DIFF
--- a/Keas.Mvc/Models/ConfirmUpdateModel.cs
+++ b/Keas.Mvc/Models/ConfirmUpdateModel.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Keas.Core.Data;
+using Keas.Core.Domain;
+using Microsoft.EntityFrameworkCore;
+
+namespace Keas.Mvc.Models
+{
+    public class ConfirmUpdateModel
+    {
+        public List<Key> Keys { get; set; }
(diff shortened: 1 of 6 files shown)


Keas.Mvc/Models/ConfirmUpdateModel.cs
```diff
@@ -0,0 +1,28 @@
...
+        public List<Key> Keys { get; set; }
+        public List<Equipment> Equipment { get; set; }
+
+
+        public static async Task<ConfirmUpdateModel> Create(string teamName, ApplicationDbContext context, User user)
+        {
+            var viewModel = new ConfirmUpdateModel
+            {
+                Keys = await context.Keys.Include(k => k.Assignment).Where(k => !k.Assignment.IsConfirmed && k.Assignment.Person.User == user && k.Team.Name == teamName).ToListAsync(),
+                Equipment = await context.Equipment.Include(e => e.Assignment).Where(e => !e.Assignment.IsConfirmed && e.Assignment.Person.User == user && e.Team.Name == teamName).ToListAsync()
```
srkirkland: @jcubbage something interesting to think about here -- User+Team is essentially Person, so we really should be able to just say where e.Assignment.Person == person.  Meaning we should start thinking about how the security service can get the person instead of the user, since we'll often want to know who the current person is instead of the user.  There should never be >1 person with the same user+team, so I think that should work easily

PR #57, opened 2018-05-10
Lh/view history

laholstege:
renamed History.Person to History.Target to make it clearer (and because scott said to)
added TargetId to Assign/Unassign in HistoryService. I wasn't sure about if Accepted should also keep track of it, so I left it off for now
equipment/keys details modal, and person page now have a history container with most recent 5 things to happen

### DIFF
--- a/Keas.Mvc/ClientApp/components/History/HistoryContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/History/HistoryContainer.tsx
@@ -1,0 +1,56 @@
+import PropTypes from "prop-types";
+import * as React from "react";
+
+import { AppContext, IHistory, IPerson } from "../../Types";
+
+import HistoryList from "./HistoryList";
+
+interface IState {
+  loading: boolean;
+  histories: IHistory[];
+}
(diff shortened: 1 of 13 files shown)


jcubbage:
Re: accept target- do we want to have "Key Accepted by James Cubbage" be one of the items I see in my history from the people controller? That should be the deciding factor here.

Keas.Mvc/ClientApp/components/History/HistoryContainer.tsx
```diff
@@ -0,0 +1,78 @@
...
+              Reload History {" "}
+              {this.state.reloaded ? <i className="fa fa-check" /> : null}
+              {this.state.reloading ? <i className="fa fa-spin fa-circle-o-notch" /> : null}
+            </Button> }
+        </div>
+      </div>
+    );
+  }
+
+  private _loadHistories = async () => {
```
srkirkland: @laholstege you could probably just reuse this in the ctor instead of doing the fetch URL twice

PR #58, opened 2018-05-10
Jsc/admin roles

jcubbage:
Add admin roles, add policy, update to handler.

### DIFF
--- a/Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
+++ b/Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
@@ -28,14 +28,24 @@ namespace Keas.Mvc.Handlers
             string team = "";
             if (context.Resource is AuthorizationFilterContext mvcContext)
             {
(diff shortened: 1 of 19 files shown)


Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -28,18 +28,28 @@ public VerifyRoleAccessHandler(ApplicationDbContext dbContext, ISecurityService
             string team = "";
```
srkirkland: @jcubbage this is fine but I prefer `string.empty`

jcubbage: So that role doesn't show up on the TeamRole list. We don't want systemAdmins being added to teams.

srkirkland: @jcubbage Ah that makes sense

PR #59, opened 2018-05-11
Jsc/team members

jcubbage:
Team member CRUD

### DIFF
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Ietws;
 using Keas.Core.Domain;
 using Keas.Mvc.Models;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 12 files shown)


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -37,5 +40,37 @@ public IdentityService(IOptions<AuthSettings> authSettings)
...
             }
         }
+
+        public async Task<User> GetUser(string id, string email)
+        {
+            // return info for the user identified by this email address in IAM
+            using (var client = new HttpClient())
+            {
+                
+                client.BaseAddress = new Uri("https://iet-ws.ucdavis.edu/api/iam/people/");
```
srkirkland: @jcubbage why manually grab it instead of using the IET webservice SDK?

jcubbage: Cause I didn't that was a thing! Should we update GetUserID to use that as well? That code is from you 8 months ago.

srkirkland: @jcubbage ha, that's funny i guess this project was the inspiration for me going out and writing the SDK.  Ok, let's leave this as is, we can go back and change it later.  BTW the SDK is https://github.com/ucdavis/ietws

PR #60, opened 2018-05-16
Jsc/workstations

jcubbage:
Workstations set-up for Laura. I also decorated all the controllers with policy. Got an error on the Asset controller, so I left that one off. Not sure what's going on. It only returns Team, so may not need security?!?!?!?!?!

### DIFF
--- a/Keas.Core/Domain/WorkstationAttribute.cs
+++ b/Keas.Core/Domain/WorkstationAttribute.cs
@@ -1,0 +1,14 @@
+using Newtonsoft.Json;
+
+namespace Keas.Core.Domain
+{
+    public class WorkstationAttribute {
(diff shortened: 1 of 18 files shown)


Keas.Core/Domain/WorkstationAttribute.cs
```diff
@@ -0,0 +1,14 @@
+using Newtonsoft.Json;
+
+namespace Keas.Core.Domain
+{
+    public class WorkstationAttribute {
+        public int Id { get; set; }
+        [JsonIgnore]
```
srkirkland: @jcubbage I think laura fixed it so we don't need this anymore?  Worth looking into later when we start actually using this part of the API

PR #62, opened 2018-05-17
Lh/edit

laholstege:
still needs auth for who can edit.
rooms aren't properly dealt with yet for keys
i also added autofocus for the assign modal


laholstege:
right now there's some issues with editing after i merged in master, but it's fixed by #61

PR #65, opened 2018-05-23
WIP: Permissions Context

srkirkland:
@jcubbage Start of the permissions context -- I created a new method so the security service can tell you your roles, and then injected them into the React context.  So now, if you go into any of the containers in React, you can add in the "permissions" context and you'll get back an array of roles, similar to `["KeyMaster","EquipMaster","AccessMaster"]` or whatever the current user's permissions are. 

Ex:

```
public static contextTypes = {
    fetch: PropTypes.func,
    permissions: PropTypes.array, // add this line if you are going to need the permissions
    router: PropTypes.object,
    team: PropTypes.object
  };
```

then use it in your component like

`if (this.context.permissions.includes('KeyMaster')) { ... }`

Feel free to create a util to offload the role checking since that might get a little messy with the admin role and everything.

### DIFF
--- a/Keas.Mvc/ClientApp/App.tsx
+++ b/Keas.Mvc/ClientApp/App.tsx
@@ -8,18 +8,21 @@ import { History } from "history";
 
 interface IProps {
   team: ITeam;
+  permissions: string[];
 }
 
 // Provider
 export default class App extends React.Component<IProps, {}> {
   public static childContextTypes = {
     fetch: PropTypes.func,
-    team: PropTypes.object
+    permissions: PropTypes.array,
+    team: PropTypes.object,
   };
   public getChildContext() {
     // define context here
     return {
       fetch: createFetch(),
+      permissions: this.props.permissions,
       team: this.props.team
     };
   }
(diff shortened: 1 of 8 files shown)


jcubbage:
How's this look? Need to add TS to my list of things to learn more of now!

Keas.Mvc/Keas.Mvc.csproj
```diff
@@ -5,6 +5,9 @@
    <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
     <TypeScriptToolsVersion>Latest</TypeScriptToolsVersion>  
   </PropertyGroup>
+  <ItemGroup>
+    <None Remove="ClientApp\util\permissions.ts" />
```
srkirkland: @jcubbage hmm, this shouldn't be here, and actually i don't think there should be any csproj changes needed.  I'm especially worried about the typescript compile item group farther down in this file -- none of the other *.ts files need this, webpack is going to do the build

jcubbage:
Reverted csproj changes and runs fine.

srkirkland:
@jcubbage 🚢

PR #68, opened 2018-06-06
Jsc/tags crud

jcubbage:


Keas.Mvc/Controllers/TagsController.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            if (ModelState.IsValid)
+            {
+                var team = await _context.Teams.FirstAsync(t => t.Name == Team);
+                newTag.Team = team;
+                _context.TeamTags.Add(newTag);
+                await _context.SaveChangesAsync();
+                Message = "Tag created.";
+                return RedirectToAction(nameof(Index));
+            }
```
jSylvestre: @jcubbage Have an ErrorMessage or similar if it isn't valid?

Keas.Mvc/Controllers/TagsController.cs
```diff
@@ -0,0 +1,113 @@
...
+            if (tag == null)
+            {
+                return NotFound();
+            }
+            return View(tag);
+        }
+
+        // POST: Tags/Edit/5
+        [HttpPost]
+        [ValidateAntiForgeryToken]
```
jSylvestre: @jcubbage FYI, there is a [AutoValidateAntiforgeryToken] you can put on the supperController so you don't need to add this attribute.

Keas.Mvc/Controllers/TagsController.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                try
+                {
+                    await _context.SaveChangesAsync();
+                    Message = "Tag updated.";
+                    return RedirectToAction(nameof(Index));
+                }
+                catch
+                {
+                    ModelState.AddModelError("", "Unable to save changes.");
```
jSylvestre: @jcubbage Did you test that this shows up? It isn't linked to a field so I think you would need the validation summary on the view.

jcubbage: Sure does. If you move this before the if, you get the page back with the error displayed.

Keas.Core/Domain/TeamTag.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Keas.Core.Domain
+{
+    public class TeamTag
```
srkirkland: @jcubbage I think I'd like it much more if this was just called Tag.  I don't think there is going to be another class of tags w/o teams, and even if there was we can just make team optional in that case

Keas.Mvc/Controllers/SuperController.cs
```diff
@@ -5,6 +5,7 @@
 namespace Keas.Mvc.Controllers
 {
     [Authorize]
+    [AutoValidateAntiforgeryToken]
```
srkirkland: @jcubbage please check that this doesn't break all of our APIs.  I don't think we are passing tokens along w/ API calls and if not this would break everything.  Give the asset pages a once over before you merge

Keas.Mvc/Controllers/TagsController.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                return NotFound();
+            }
+            return View(tag);
+        }
+
+        // POST: Tags/Delete/5
+        [HttpPost]
+        public async Task<ActionResult> Delete(int id, TeamTag deleteTag)
+        {
```
srkirkland: @jcubbage good security paranoia would be to make sure the Id of the tag to delete is an existing tag which belongs to the current team before updating/deleting.  Otherwise anyone can delete the tag of any other team just by posting random info

Keas.Core/Domain/TeamTag.cs
jcubbage: @srkirkland So, I started by calling it Tag, but that conflicted with the property of Tag. I couldn't come up with a good alternative other than TeamTag.Tag. Thoughts?

Keas.Mvc/Controllers/SuperController.cs
jSylvestre: @srkirkland Just use [IgnoreAntiforgeryTokenAttribute] where needed?

jcubbage: <in Kirk screaming Kahn voice>JJJJJJJAAAAAASSSSSOOOONNNN!!!!!</voice>

jcubbage: Totally broke all the React API calls.

jSylvestre: @jcubbage 
An example if it helps later on:

```const antiforgery = $("input[name='__RequestVerificationToken']").val();
    $.post({
      url: postUrl,
      data: { model: order, __RequestVerificationToken: antiforgery }
    })```

PR #72, opened 2018-06-07
move to cas keyed off kerb instead of AzureAD

srkirkland:


jcubbage:
My testsecurity pages all respond accordingly as I adjust my roles. Looks good.

PR #74, opened 2018-06-14
Lh/search tags

laholstege:


Keas.Mvc/Controllers/TagsController.cs
```diff
@@ -11,7 +11,6 @@
 
 namespace Keas.Mvc.Controllers
 {
-    [Authorize(Policy = "DepartmentAdminAccess")]
```
jcubbage: I think it's better to move this back up to cover the whole controller, and then add [AllowAnonymous] on the listTags action. That way if we later change the policy to something else (say TagMaster), we only have to change in one place.

Keas.Mvc/ClientApp/components/Tags/SearchTags.tsx
```diff
@@ -0,0 +1,34 @@
...
+    public render() {
+        return (
+            <div>
+                <Typeahead
+                    options={this.props.disabled ? [] : this.props.tags}
+                    disabled={this.props.disabled}
+                    multiple={true}
+                    clearButton={true}
+                    onChange={selected => {
+                        this.props.onSelect(selected);
```
srkirkland: @laholstege if you are going to call onSelect() here, then it should be a require prop.  Is there a scenario in which we'd want it optional like the IProps interface would suggest?

Keas.Mvc/Controllers/TagsController.cs
srkirkland: @laholstege Well I don't want to make it truly [AllowAnonymous] because only department users should be able to access this, but I also agree that we should leave the authorize policy as-in on the main controller in here.  Probably the way to go is just to move the listTags into a new controller, or maybe into the assetcontrolelr

Keas.Mvc/Startup.cs
```diff
@@ -142,7 +142,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                     name: "API",
                     template: "api/{teamName}/{controller}/{action}/{id?}",
                     defaults: new { controller = "people", action = "Index" },
-                    constraints: new { controller = "(keys|equipment|access|spaces|people|person|workstations)" }
+                    constraints: new { controller = "(keys|equipment|access|spaces|people|person|workstations|tags)" }
```
srkirkland: @laholstege I worry that this will mess up the existing tags controller, so I'm definitely thinking now we just move the tag lookup part to some new controller so we don't get conflicts

PR #75, opened 2018-07-24
Jcs/email job

jSylvestre:


Keas.Jobs.Core/Class1.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace Keas.Jobs.Core
+{
+    public class Class1
+    {
+    }
+}
```
jcubbage: Is this class needed? I don't see it used, and the name is suspicious or not actually being used..

jSylvestre: @jcubbage No, but if we add Serilog, it will be replaced with a log helper class. This was just the default that got created when I created the job. I'll delete it.

Keas.Jobs.SendMail/Program.cs
```diff
@@ -0,0 +1,65 @@
...
+
+
+            // Use this to get configuration info, environmental comes in from azure
+            var builder = new ConfigurationBuilder()
+                .SetBasePath(Directory.GetCurrentDirectory())
+                .AddJsonFile("appsettings.json");
+            var environmentName = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
+            if (string.Equals(environmentName, "development", StringComparison.OrdinalIgnoreCase))
+            {
+                builder.AddUserSecrets<Program>();
```
jSylvestre: @srkirkland I don't know if there is a better/updated way to do this, but this is how I did it with anlab

srkirkland: @jSylvestre yes that is old and you should not need that anymore

PR #76, opened 2018-07-25
Jcs/fix db init

jSylvestre:


jSylvestre:
@srkirkland yes, I deleted the keas.db file and ran it and it looked ok to me. 
If you or @jcubbage wants to double check and test it too it wouldn't hurt.

PR #77, opened 2018-08-06
Lh/react table

laholstege:


Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -94,13 +100,56 @@ export default class EquipmentContainer extends React.Component<IProps, IState>
...
+          onAdd={this._openAssignModal}
+          showDetails={this._openDetailsModal}
+          onEdit={this._openEditModal}
+        />
+    </div>);
+    }
+    else
+    {
+      let filteredEquipment = this.state.equipment;
+      if(!!this.state.tagFilters && this.state.tagFilters.length > 0)
```
srkirkland: @laholstege FYI I don't think the `!!` is doing anything here -- an array is truthy if it exists and shouldn't need the coalescence that a string would require.  However it doesn't hurt anything either, just makes it a very small bit harder to read.

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -195,6 +244,37 @@ export default class EquipmentContainer extends React.Component<IProps, IState>
       equipment: updateEquipment
     }); 
   }
+
+  private _filterTags = (filters: string[]) => {
+    this.setState({tagFilters: filters});
+}
+
+  private _checkTagFilters = (equipment: IEquipment, filters: string[]) => {
```
srkirkland: @laholstege you can probably write this in a more succinct way with .Every() if I understand the logic here.  Something like `filters.every(f => equipment.tags.indexOf(filter) !== -1)` should do it.  Or even get fancy perhaps and go `filters.every(f => equipment.tags.contains(filter))`

Keas.Mvc/ClientApp/components/Equipment/EquipmentTable.tsx
```diff
@@ -0,0 +1,110 @@
...
+        data={this.props.equipment}
+        filterable={true}
+        minRows={1}
+        columns = {[
+            {
+                Header: "Serial Number",
+                accessor: "serialNumber",
+                filterMethod: (filter, row) => 
+                    !!row[filter.id] &&
+                    row[filter.id].toLowerCase().indexOf(filter.value) !== -1,
```
srkirkland: @laholstege personal preference but I like .contains() instead of indexOf so we don't have remember what -1 means.  Also again I don't think the `!!` is doing anything

Keas.Mvc/ClientApp/components/Equipment/EquipmentTable.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    const now = moment();
+                    if( filter.value === "expired") {
+                        return !!row._original.assignment && moment(row._original.assignment.expiresAt).isSameOrBefore(moment(), "day")
+                    }
+                    if( filter.value === "unexpired") {
+                        return !!row._original.assignment && moment(row._original.assignment.expiresAt).isAfter()
+                    }
+                    if( filter.value === "3weeks") {
+                        return !!row._original.assignment && moment(row._original.assignment.expiresAt).isAfter() 
+                            && moment(row._original.assignment.expiresAt).isBefore(moment().add(3,'w'))
```
srkirkland: cool!

Keas.Mvc/ClientApp/components/Spaces/SpacesList.tsx
```diff
@@ -11,38 +12,92 @@ interface IProps {
...
+                data={this.props.spaces}
+                filterable={true}
+                minRows={1}
+                columns = {[
+                    {
+                        Header: "Room",
+                        accessor: (row) => row.space.roomNumber + " " + row.space.bldgName,
+                        id: "room",
+                        filterMethod: (filter, row) => {
+                            return row[filter.id].toLowerCase().indexOf(filter.value.toLowerCase()) !== -1;
```
srkirkland: Double checking here but do you need to check existence of `row[filter.id]`?  Seems like everywhere else in the PR you are doing a check before accessing .toLowerCase() but not in this one

laholstege: @srkirkland this should always exist, whereas "Room Name" isn't required--but I added it just to be safe

PR #81, opened 2018-08-09
Lh/people and spaces tables

laholstege:
people/spaces table now only have "view details" action that directs to a different view, instead of being a modal popup. 

can now do all actions directly from these new views. these actions call props passed in to update the counts shown in the table. 

filters are also now controlled by the Container, e.g. SpacesContainer controls both tagFilters and tableFilters so they aren't lost when you navigate to the details 

still have to pass down space instead of spaceID and use when assigning/creating from spaces/details


jcubbage:
Anyone else lose the menu on the ellipse on the Equipment tab?  When I click the blue ..., nothing happens, but only on the equipment tab. Equipment menu works when looking at spaces or people.

laholstege:
@jcubbage hmm, it's working fine for me. have you done `npm install`? only thing i can think of

jcubbage:
That didn't fix it. Reboot didn't fix. When I open the React tools in Chrome, I see the isOpen on the dropdown toggle, and the open class appear/disappear from the div and dropdownToggle. I also see the dropdownitems appear and disappear from the dropdownmenu item. But nothing visibly appears on the screen. Only on that one page.....sigh.

jSylvestre:
@jcubbage Did the database change, and if it did, are you using sql express? Just a random guess.

Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
```diff
@@ -148,6 +156,10 @@ export default class AccessContainer extends React.Component<IProps, IState> {
             access: [...this.state.access, access]
         });
     }
+    if(!!this.props.accessAssigned)
```
srkirkland: @laholstege No need for the `!!` here since this is a function and doesn't need the coalescence.  Just a regular `if(this.props.accessAssigned)` will serve the same purpose

Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
```diff
@@ -18,33 +24,262 @@ export default class PeopleContainer extends React.Component<{}, IState> {
...
+    //     equipment: updateEquipment
+    //   });
+    // } else {
+    //   this.setState({
+    //     equipment: [...this.state.equipment, equipment]
+    //   });
+    // }
+  };
+
+  private _revokeEquipment = async (equipment: IPerson) => {
```
srkirkland: @laholstege what's all these commented out methods about?  Is it because we are using the underlying containers to do all the work now and these are no longer needed?  If so I say just dump them if you don't need it any more.

Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
```diff
@@ -18,33 +24,262 @@ export default class PeopleContainer extends React.Component<{}, IState> {
...
+            people[index].accessCount++;
+            break;
+          case "workstation":
+            people[index].workstationCount++
+        }
+        this.setState({people});
+    } 
+}
+
+private _assetRevoked = (type: string, spaceId: number, personId: number) => {
```
srkirkland: @laholstege probably you can combine these too methods to keep it DRY.  Maybe _assetCountUpdated() where you pass along the personId and the delta and then just do += math

Keas.Mvc/ClientApp/components/Spaces/SpacesContainer.tsx
```diff
@@ -125,38 +125,53 @@ export default class SpacesContainer extends React.Component<{}, IState> {
...
         } 
     }
 
-    private _workstationRevoked = (spaceId: number) => {
-        const index = this.state.spaces.findIndex(x => x.id === spaceId);
-        if(index > -1)
-        {
-            const spaces = [...this.state.spaces];
+    private _assetRevokedOrDeleted = (type: string, spaceId: number, personId: number) => {
+    const index = this.state.spaces.findIndex(x => x.id === spaceId);
```
srkirkland: @laholstege seems like this code is pretty common across multiple files,  Might not be worth it but think about moving to a common util method you can just import into the page.  Something where you'd pass the list, the index you care about, the delta, and the type and let the method figure it out.

srkirkland:
@jcubbage I didn't have that issue but there are some odd style issues with the tables for sure so I'm not that surprised.  What browser are you using and at what resolution?  We're going to have Cal look through and it'll help if he can reproduce the weirdness.

jcubbage:
I had the same issue on Chrome (Version 68.0.3440.106 (Official Build) (64-bit)) and Firefox (Extended Support Release 60.1.0esr (32-bit). Browser was running in 1080x1920 portrait mode, though moving to same resolution on a landscape didn't fix it. Both browsers showed nothing when I clicked the button (but ONLY on the equipment tab). Surface is running 2736x1824, Chrome (67.0.3396.99 (64bit), Firefox  (61.0.2 (64-bit).  Surface had some of the context menu appear, but it was cut-off.

laholstege:
I had the same problem earlier with the button dropdown not showing up, but doing `npm install` fixed it for me

jcubbage:
Works fine when I install VSCode and run it.

PR #82, opened 2018-08-10
Jsc/new keys

jcubbage:


Keas.Core/Domain/Key.cs
```diff
@@ -1,18 +1,13 @@
...
-        public string SerialNumber { get; set; }
-
-        public string LockId { get; set; }
-        public Space Space { get; set; }
-        public int? SpaceId { get; set; }
-
-        public KeyAssignment Assignment { get; set; }
-        public int? KeyAssignmentId { get; set; }
+        public List<KeyXSpace> KeyXSpaces { get; set; }
+        public List<Serial> Serials { get; set; }
```
laholstege: @jcubbage @srkirkland should we keep a list of serials here, or should we just have serials point to a key? like spaces and workstations?
right now I'm getting nested stuff in the react, because each serial has its key, and then that key has a list of serials. I'm thinking either we pass in keys to the list, or we remove the list of serials here and group serials by KeyId

jcubbage: Pull the list out or mark it not for JSON serialization. I put it there thinking it would help with the queries. But I ended up making serials the base of my queries, so I don't think it got used anywhere.

Keas.Core/Domain/History.cs
```diff
@@ -28,8 +28,13 @@ public History()
         // Created, granted, accepted, declined, revoked, deleted 
         public string ActionType { get; set; }
 
+        // TODO: Add serial? Or just replace key with serial????
```
srkirkland: @jcubbage we probably want both anyway so remove this comment and i think the way you implemented it is a good plan

Keas.Core/Domain/KeyXSpace.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class KeyXSpace
+    {
+        [Key]
+        public int Id { get; set; }
+
+        public Key Key { get; set; }
+        public int KeyId { get; set; }
+
+        public Space Space { get; set; }
+        public int? SpaceId { get; set; }
```
srkirkland: @jcubbage why is this nullable?  Doesn't seem like there should be the possibility of having an entry in this table w/o both a keyId and spaceId.

PR #85, opened 2018-08-16
Lh/better count updates

laholstege:
counts now update better on spaces and peoples tab. 

especially on spaces tab, where you can edit an asset and reassign it to another space. it will now properly reflect that change, i.e. removing from state and updating counts for both the old and new space. 

still need to implement this for keys, but i figured it'd be cleaner just to wait for the changes on that

### DIFF
--- a/Keas.Mvc/ClientApp/components/Keys/KeyContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/KeyContainer.tsx
@@ -152,11 +152,13 @@ export default class KeyContainer extends React.Component<IProps, IState> {
     }
     if(created)
     {
-        this.props.keyTotalUpdated("key", this.props.space.id, this.props.person ? this.props.person.id : null, 1);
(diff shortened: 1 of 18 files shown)


Keas.Mvc/ClientApp/components/Keys/KeyContainer.tsx
```diff
@@ -16,11 +16,11 @@ interface IState {
 }
 
 interface IProps {
-  keyAssigned?: (type: string, spaceId: number, personId: number, created: boolean, assigned: boolean) => void;
-  keyRevoked?: (type: string, spaceId: number, personId: number) => void;
+  keyInUseUpdated: (type: string, spaceId: number, personId: number, count: number) => void;
+  keyTotalUpdated: (type: string, spaceId: number, personId: number, count: number) => void;
```
srkirkland: @laholstege why bother calling it `keyTotalUpdated` when (1) it's already in the key container, and (2) you are passing "key" as the first param anyway.  Same thing with other asset types

PR #88, opened 2018-08-22
Update for person = null and redirects need return

jcubbage:
Corrects for error Jason was seeing.

### DIFF
--- a/Keas.Mvc/Controllers/ConfirmController.cs
+++ b/Keas.Mvc/Controllers/ConfirmController.cs
@@ -25,9 +25,9 @@ namespace Keas.Mvc.Controllers
         public async Task<IActionResult> MyStuff()
         {   
             var person = await _securityService.GetPerson(Team);
-            if(person==null){
+            if(person == null){
                  Message = "You are not yet added to the system.";
-                return RedirectToAction("Index","Home");
+                return RedirectToAction("NoAccess","Home");
             }
             var viewmodel = await MyStuffListModel.Create(_context, person);
             return View(viewmodel);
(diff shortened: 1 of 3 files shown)


Keas.Mvc/Controllers/ConfirmController.cs
```diff
@@ -23,20 +23,28 @@ public ConfirmController(ApplicationDbContext context, ISecurityService _securit
         }
 
         public async Task<IActionResult> MyStuff()
-        {
+        {   
             var person = await _securityService.GetPerson(Team);
+            if(person==null){
+                 Message = "You are not yet added to the system.";
+                return RedirectToAction(nameof(HomeController.Index));
```
jSylvestre: @jcubbage Need controller name,

Keas.Mvc/Controllers/ConfirmController.cs
```diff
@@ -23,20 +23,28 @@ public ConfirmController(ApplicationDbContext context, ISecurityService _securit
         }
 
         public async Task<IActionResult> MyStuff()
-        {
+        {   
             var person = await _securityService.GetPerson(Team);
+            if(person==null){
```
srkirkland: Spaces in the operator here please

Keas.Mvc/Controllers/ConfirmController.cs
```diff
@@ -23,20 +23,28 @@ public ConfirmController(ApplicationDbContext context, ISecurityService _securit
...
+                 Message = "You are not yet added to the system.";
+                return RedirectToAction("Index","Home");
```
srkirkland: Interesting, we probably want to make this a much nicer experience in the future, maybe a page with info on what to do.

jcubbage:
Better?

PR #89, opened 2018-08-23
Lh/person

laholstege:
create/edit person

bio container still needs to show more info 
not sure if we want to let people edit active from here


Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
```diff
@@ -141,6 +150,66 @@ export default class PeopleContainer extends React.Component<{}, IState> {
...
+    {
+      // if we somehow already have this person, return
+      return;
+    }
+    person.teamId = this.context.team.id;
+    person = await this.context.fetch(`/api/${this.context.team.name}/people/create`, {
+      body: JSON.stringify(person),
+      method: "POST"
+      })
+      .catch(err => {
```
srkirkland: @laholstege With async/await instead of using .catch() you should just wrap a try/catch statement.   .catch() is for when you are using the old promise.then().catch() style.

Keas.Mvc/ClientApp/components/People/SearchUsers.tsx
```diff
@@ -0,0 +1,64 @@
...
+        </div>
+        );
+  }
+
+  private _loadUser = async () => {
+    this.setState({reloading: true, reloaded: false});
+    const userFetchUrl = `/api/${this.context.team.name}/people/searchUser?searchTerm=${this.state.search}`;
+  
+    var person = await this.context.fetch(userFetchUrl)
+      .catch(err => {
```
srkirkland: try/catch

PR #96, opened 2018-08-24
Testing message

jSylvestre:


jcubbage:
Time to drink the cool-aid and switch to VS Code!

PR #105, opened 2018-09-14
Jsc/reports

jcubbage:
I think this is ready to go (except the workstation bit. Will need to get EmailJob PR approved first).

I know we had a user story for expiring items, which this completes. We also had one for searching for keys with a bunch of parameters, but I think the React pages do all that. Maybe add some export for that?

### DIFF
--- a/Keas.Mvc/Keas.Mvc.csproj
+++ b/Keas.Mvc/Keas.Mvc.csproj
@@ -8,7 +8,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="AspNetCore.Security.CAS" Version="2.0.1" />
-    <PackageReference Include="Bootstrap.Datepicker" Version="1.8.0" />
     <PackageReference Include="ietws" Version="0.1.7" />
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0" />
--- a/Keas.Mvc/Keas.Mvc.csproj
+++ b/Keas.Mvc/Keas.Mvc.csproj
@@ -8,6 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="AspNetCore.Security.CAS" Version="2.0.1" />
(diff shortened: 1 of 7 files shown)


Keas.Mvc/Keas.Mvc.csproj
```diff
@@ -8,6 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="AspNetCore.Security.CAS" Version="2.0.1" />
+    <PackageReference Include="Bootstrap.Datepicker" Version="1.8.0" />
```
srkirkland: @jcubbage why are we installing this as a NUGET package?  Seems like it should be an NPM package like our other JS libraries (ex: bootstrap itself)

Keas.Mvc/Models/ExpiringItemsViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+        public static ExpiringItemsViewModel Create(ApplicationDbContext context, DateTime expiresBefore, string teamName, bool showInactive, string showType)
+        {
+            // AccessAssignement needs db update to link back to Access.
+            //var expiringAccess = context.AccessAssignments.Where(a => a.ExpiresAt <= expiresBefore).Include(a => a.Access).AsNoTracking();
+            //var expiringAccess = context.Access.Where(a => (showType == "All" || showType == "Access") &&
+            //                                             a.Team.Name == teamName && a.Assignment.ExpiresAt <= expiresBefore && (a.Access.Active || a.Access.Active == !showInactive))
+            //    .Include(a => a.Assignment).ThenInclude(a => a.Person).AsNoTracking();
+            var expiringKey = context.Serials.Where(a => (showType == "All" || showType == "Key") &&
+                a.Key.Team.Name == teamName && a.Assignment.ExpiresAt <= expiresBefore && (a.Key.Active || a.Key.Active == !showInactive))
+                .Include(k => k.Assignment).ThenInclude(a=> a.Person).Include(k => k.Key).AsNoTracking();
```
srkirkland: @jcubbage I think it's worth while to add on .ToList() or better yet .ToArray() here.  If we don't we risk our consuming view page actually executing the queryable or an N+1 issue.  It'd be better if the data access was done here, and as a plus we could also do them async here as well.

PR #122, opened 2018-10-01
Add Myself to db init

jSylvestre:


jcubbage:
This is going to break the other changes I had to make to DBInit!

PR #127, opened 2018-10-02
Get rid of status on emails. Don't think it's used.

jcubbage:


jSylvestre:
@jcubbage Remove from the DB too?

PR #130, opened 2018-10-03
Update role checking

jSylvestre:


srkirkland:
selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

PR #163, opened 2018-10-18
Jsc/dashboard

### LINKED ISSUE #148: Update Dashboard
For the "homepage" dashboard (i.e. https://keas.azurewebsites.net/caesdo), let's clean it up to remove testing pages.

Dashboard Should have:
Assets (if they can see their team assets)
My Stuff (ideally we can put a badge up if they have assets waiting to confirm)
Team Administration (if they are an admin)

Header Bar should have:
Assets (if they have access)
My Stuff
Reports (if they have access)
Admin (if they are admin)

jcubbage:
Fixes #148, #164 
Drop down not working fully. If you pull from cdn instead of local it works. So, I think we need to updated locals of bootstrap.js.

### DIFF
--- a/Keas.Mvc/Controllers/HomeController.cs
+++ b/Keas.Mvc/Controllers/HomeController.cs
@@ -14,6 +14,10 @@ namespace Keas.Mvc.Controllers
     {
         public IActionResult Index()
         {
+            string Team = ControllerContext.RouteData.Values["teamName"] as string;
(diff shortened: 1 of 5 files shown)


jcubbage:
Update to helpers works for everything except /teamslug/keys. Helpers want to add /api/ in front of team. I couldn't find a combo that got rid of that.

PR #165, opened 2018-10-19
Jsc/access to table2

### LINKED ISSUE #109: put access details into a table

jcubbage:
Closes #109


Keas.Mvc/ClientApp/components/Access/AccessDetails.tsx
```diff
@@ -9,22 +9,24 @@ import {
...
     selectedAccess: IAccess;
+    onRevoke: (accessAssignment: IAccessAssignment) => void;
 }
 
 
 export default class AccessDetails extends React.Component<IProps, {}> {
 
     public render() {
-        if (this.props.selectedAccess == null)
+        if (this.props.selectedAccess == null){
```
srkirkland: @jcubbage Try to avoid the double equals -- i think you could either go with `this.props.selectedAccess === null` or simply `!this.props.selectedAccess`, depending on if you really need to check for null or just a falsey value.  I always love the later but it is semantically ever so slightly different

jcubbage: So, this was actually TSLint complaining about the missing curly brackets from someone else. I'll update to triple equal.

PR #175, opened 2018-10-22
Jsc/fix expirations

### LINKED ISSUE #144: Format of expiration
Date only, no need for ISO with time

jcubbage:
Fixes #144

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
@@ -4,6 +4,7 @@ import { IEquipment, IEquipmentAttribute, ISpace } from "../../Types";
 import AssignSpace from "../Spaces/AssignSpace";
 import SearchTags from "../Tags/SearchTags";
 import EquipmentAttributes from "./EquipmentAttributes";
+import * as moment from "moment";
 
 interface IProps {
     changeProperty?: (property: string, value: any) => void;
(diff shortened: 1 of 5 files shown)


Keas.Mvc/ClientApp/components/Equipment/EquipmentListItem.tsx
```diff
@@ -15,14 +17,14 @@ interface IProps {
 export default class EquipmentListItem extends React.Component<IProps, {}> {
     public render() {
         const hasAssignment = !!this.props.equipmentEntity.assignment;
+        const dates = this.props.equipmentEntity.assignment.expiresAt;
+        const expirationDate = hasAssignment ? moment(dates).format("MM/DD/YYYY").toString() : "";
         return (
           <tr>
             <td>{this.props.equipmentEntity.serialNumber}</td>
-            <td>{this.props.equipmentEntity.name}</td>
+            <td>{this.props.equipmentEntity.name} kkkk</td>
```
srkirkland: ???

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -56,7 +57,7 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
                     <input type="text"
                         className="form-control"
                         disabled={true}
-                        value={this.props.selectedEquipment.assignment.expiresAt.toString()}
+                        value={moment(this.props.selectedEquipment.assignment.expiresAt).format("MM/DD/YYYY").toString()}
```
srkirkland: @jcubbage how about instead you make a quick util (kinda like the permissions util) and just have it export a method like formatExpiration() that takes a date.  It can handle the null checks, importing moment and defining the format.  Should make the code cleaner plus if we want to change the format slightly later we just do it in one place

jcubbage:
Switching to Util instead of on each page.

PR #182, opened 2018-10-23
Jsc/better my stuff

### LINKED ISSUE #169: Alert message if you have pending acceptance tasks
Would be nice if on the "MyStuff" page and maybe even the homepage, if you have something pending acceptance you get a big message telling you to click it to go and accept stuff

jcubbage:
Fixes #169

### DIFF
--- a/Keas.Mvc/Controllers/HomeController.cs
+++ b/Keas.Mvc/Controllers/HomeController.cs
@@ -6,19 +6,38 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Keas.Mvc.Models;
 using Microsoft.AspNetCore.Authorization;
+using Keas.Core.Data;
+using Keas.Mvc.Services;
 
 namespace Keas.Mvc.Controllers
 {
     [Authorize]
     public class HomeController : SuperController
     {
+        private readonly ApplicationDbContext _context;
+
+         private readonly ISecurityService _securityService;
+
+       
+        public HomeController(ApplicationDbContext context, ISecurityService securityService)
+        {
+            _context = context;
+            _securityService = securityService;
+        }
+
         
-        public IActionResult Index()
+        public async Task<ActionResult> Index()
         {            
             if (Team == null){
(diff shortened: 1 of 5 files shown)


Keas.Mvc/Controllers/HomeController.cs
```diff
@@ -6,19 +6,38 @@
...
 using Microsoft.AspNetCore.Authorization;
+using Keas.Core.Data;
+using Keas.Mvc.Services;
 
 namespace Keas.Mvc.Controllers
 {
     [Authorize]
     public class HomeController : SuperController
     {
+        private readonly ApplicationDbContext _context;
```
srkirkland: @jcubbage the formatting here is kinda strange all over the file.  The private vars aren't aligned, the if statement has no parenthesis spacing and the content inside the if statement isn't aligned either.  If this is in VS code maybe make sure you have the C# code extension installed and run the formatter

Keas.Mvc/Models/HomeViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public class HomeViewModel
+    {
+        public bool PendingItems { get; set; }
+
+
+
+        public static async Task<HomeViewModel> Create(ApplicationDbContext context, Person person)
+        {   
+            var viewModel = new HomeViewModel();
```
srkirkland: I don't love doing 3 queries on the homepage but at least it's pretty simple.  Nothing to worry about but let's keep an eye on the performance and make sure it doesn't tax the azure db

Keas.Mvc/Controllers/HomeController.cs
jcubbage: Better?

Keas.Mvc/Models/HomeViewModel.cs
jcubbage: Yeah...I wasn't super thrilled about putting this here. My assumption is at a future date we do this once for a user and set some session flag that gets checked every 5 minutes or so in Ver 2.0.

PR #185, opened 2018-10-23
Lh/extend assignments

### LINKED ISSUE #166: Ability to edit/extend expiration date on any asset assignment.

laholstege:
closes #166
still working on adding it for access, since that's more complicated.

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -173,8 +173,8 @@ export default class EquipmentContainer extends React.Component<IProps, IState> 
     equipment: IEquipment,
     date: any
   ) => {
-    let created = false;
-    let assigned = false;
+    let updateTotalAssetCount = false;
+    let updateInUseAssetCount = false;
 
     const attributes = equipment.attributes;
     // call API to create a equipment, then assign it if there is a person to assign to
@@ -186,7 +186,7 @@ export default class EquipmentContainer extends React.Component<IProps, IState> 
         method: "POST"
       });
       equipment.attributes = attributes;
-      created = true;
+      updateTotalAssetCount = true;
     }
 
     // if we know who to assign it to, do it now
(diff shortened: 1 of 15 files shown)


jcubbage:
Was able to extend on the equipment from the Space tab. But, I'm getting an error on the equipment tab: EquipmentTable.tsx:51 Uncaught TypeError: Cannot read property 'expiresAt' of null
    at Object.accessor (EquipmentTable.tsx:51)

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -195,12 +195,16 @@ export default class EquipmentContainer extends React.Component<IProps, IState>
         person.id
       }&date=${date}`;
 
+      if(!equipment.assignment)
+      {
+        // don't count as assigning unless this is a new one
+        assigned = true;
```
srkirkland: @laholstege seems a bit odd, why are we setting assigned = true if there is no equipment assignment? Is it just the variable naming?

Keas.Mvc/ClientApp/components/Keys/AssignKey.tsx
```diff
@@ -45,10 +45,11 @@ export default class AssignKey extends React.Component<IProps, IState> {
   constructor(props) {
     super(props);
     this.state = {
-      date: moment().add(3, "y"),
-      error: "",
+      date: (!!this.props.selectedKey && !!this.props.selectedKey.assignment)
```
srkirkland: Let's think about adding some helpers in the future that can determine if an assignment exists and to set the default expiration date.  Something like if `hasAssignment(this.props.selectedKey) { }` and `getDefaultExpiration()`, etc, etc.

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
laholstege: @srkirkland you're right, it's kind of confusing. the variable is used to determine if the function for updating counts should be called. so it should only be true if this is a new assignment.

srkirkland: @laholstege so it's more like a shouldUpdate or hasChanges or similar?  I have to make a big deal about variables because if I'm confused now i'll really be confused reading this again in 3 years

srkirkland:
@laholstege so pretty!selfie-0

selfie-1
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

PR #186, opened 2018-10-23
Update to avoid null check in Equipment Assignment

jcubbage:
Is this any better?

### DIFF
--- a/Keas.Mvc/ClientApp/util/dates.ts
+++ b/Keas.Mvc/ClientApp/util/dates.ts
@@ -1,5 +1,5 @@
 import * as moment from "moment";
-import { IEquipmentAssignment} from "../Types";
+import { IHasExpiration} from "../Types";
 
 export class DateUtil {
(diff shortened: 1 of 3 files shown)


jcubbage:
@srkirkland Did you see the updates?

srkirkland:
@jcubbage not yet let me look

Keas.Mvc/ClientApp/util/dates.ts
```diff
@@ -9,6 +10,16 @@ export class DateUtil {
         return moment(expiration).format("MM/DD/YYYY").toString();
     }
 
+    public static formatAssignmentExpiration(assignment: IHasExpiration) {
+        if(assignment === null) {
```
srkirkland: @jcubbage You can simplify to `if (!assignment || !assignment.expiresAt) {}` but otherwise I like this I think it's a clean way of doing it

PR #187, opened 2018-10-23
add active tag

### LINKED ISSUE #184: make sure queries check Active flag

laholstege:
closes #184

### DIFF
--- a/Keas.Mvc/Controllers/Api/AccessController.cs
+++ b/Keas.Mvc/Controllers/Api/AccessController.cs
@@ -51,7 +51,7 @@ namespace Keas.Mvc.Controllers.Api
         public async Task<IActionResult> List()
         {
             var accessList = await _context.Access
-                .Where(x => x.Team.Slug == Team)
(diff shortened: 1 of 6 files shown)


jcubbage:
Should this be a flag on the react components? How do we pull up inactive items (in case we need to reactivate). Also, we need some way of setting this flag (although I can be convinced that's a separate issue/PR).

laholstege:
@jcubbage i think react will only deal with active items, except maybe to de-activate/remove assets. we would have some static page for reactivating. what do you think, @srkirkland ?

srkirkland:
@laholstege @jcubbage yea we'll just have an inactive report somewhere.  I'm not sure we should ever re-activate an old thing except maybe people.

srkirkland:
@laholstege something to think about -- we can set the class mapping so that isActive is automatically applied at the model level, then you don't have to add it in to each query.

srkirkland:
@laholstege that would be like https://docs.microsoft.com/en-us/ef/core/querying/filters

PR #191, opened 2018-10-24
Jsc/fix new team

jcubbage:
Fixes bug mentioned in slack channel.

### DIFF
--- a/Keas.Mvc/Controllers/TeamController.cs
+++ b/Keas.Mvc/Controllers/TeamController.cs
@@ -41,7 +41,7 @@ namespace Keas.Mvc.Controllers
             {
                 _context.Add(team);
                 await _context.SaveChangesAsync();
-                return RedirectToAction("Index","Admin");
+                return RedirectToAction("Index","Admin", "Please");
             }
             return View(team);
         }
--- a/Keas.Mvc/Controllers/TeamController.cs
+++ b/Keas.Mvc/Controllers/TeamController.cs
@@ -41,7 +41,7 @@ namespace Keas.Mvc.Controllers
             {
                 _context.Add(team);
                 await _context.SaveChangesAsync();
-                return RedirectToAction("Index","Admin", "Please");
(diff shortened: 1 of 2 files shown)


Keas.Mvc/Controllers/TeamController.cs
```diff
@@ -41,7 +41,7 @@ public async Task<IActionResult> Create([Bind("Id,Name,Slug")] Team team)
             {
                 _context.Add(team);
                 await _context.SaveChangesAsync();
-                return RedirectToAction("Index","Admin");
+                return RedirectToAction("Index","Admin", "new {id=1}");
```
srkirkland: hmm strange, what if there is no team 1?  And why do we want to redirect them there anyway?

jSylvestre: @srkirkland The id isn't used. It is just the index page.
@jcubbage Same issue with the back to list link on the create team page

srkirkland: @jSylvestre is this because there is something wrong w/ the slug?  why aren't we using that

jSylvestre: @srkirkland TeamController doesn't use slugs.  Looks like SysAdmin page:
[Authorize(Policy = "SystemAdminAccess")]

jcubbage: For some reason, the redirect goes to /Admin/Home....which is Home controller with a Teamslug = "Admin". Since that team doesn't exist, you get a redirect to no access. Actually, after my other PR, it will be fine, but you're on a team that doesn't exist (until someone makes a team Admin).

jcubbage: By giving it a id in the redirect, it properly redirects to site/Admin/Index page.

jcubbage:
Fixed via new route in Pr #202

PR #196, opened 2018-10-24
Jsc/role gets personhood

### LINKED ISSUE #194: Only Departmental Admin Access
This is all I see if I only have DA access.
Note, It does pick the correct team automatically.

image

jcubbage:
Fixes #194 Now if you create a roled user, they will also have personhood established.  
Will need to recreate role to fix the page you saw.

### DIFF
--- a/Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
+++ b/Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
@@ -9,6 +9,8 @@ using Keas.Mvc.Services;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Microsoft.AspNetCore.Http;
 
 namespace Keas.Mvc.Handlers
 {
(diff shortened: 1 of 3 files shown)


Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -32,11 +39,16 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
                 {
                     team = mvcContext.RouteData.Values["teamName"].ToString();
                 }
-            }
+            }  
+            if (string.IsNullOrWhiteSpace(team)) 
+             {
+                var tempData = _tempDataDictionaryFactory.GetTempData(_httpContext.HttpContext);
+                team = tempData["TeamName"] != null ? tempData["TeamName"].ToString() : "";
```
srkirkland: @jcubbage why not just do team = `string.format("{0}", tempData["TeamName"})` or maybe even simpler use `Convert.ToString(tempData["TeamName"]`.  That'll take care of nulls and formatting as a string

Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -32,11 +39,16 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
+            if (string.IsNullOrWhiteSpace(team)) 
+             {
+                var tempData = _tempDataDictionaryFactory.GetTempData(_httpContext.HttpContext);
+                team = tempData["TeamName"] != null ? tempData["TeamName"].ToString() : ""; 
+            }     
 
             var user = _dbContext.Users.SingleOrDefault(u => u.Id == context.User.Identity.Name);
             var roles = await _dbContext.Roles.Where(r => requirement.RoleStrings.Contains(r.Name)).ToListAsync();
-            if (user != null && team!="")
+            if (user != null && team != "")
```
srkirkland: @jcubbage use string.IsNullOrEmpty() here instead of comparing to "".

PR #199, opened 2018-10-25
Use team slug instead of OrgId.

### LINKED ISSUE #198: All spaces currently showing up for all teams.
All calls to /api/space/list have the same orgID of "ADNO".
SpaceContainer, Ln 39: 
const spaces = await this.context.fetch(`/api/${this.context.team.slug}/spaces/list?orgId=ADNO`);

jcubbage:
Fixes #198 
Only uses orgcode, ignores chart. Not sure if that's an issue. At least in dev, no chart in space table.

### DIFF
--- a/Keas.Core/Domain/Space.cs
+++ b/Keas.Core/Domain/Space.cs
@@ -5,7 +5,7 @@ namespace Keas.Core.Domain
 {
     // Facilities link deptSpace, flattened
     // ties together rooms and departments
-    [Table("Spaces")]
+    [Table("Space")]
     public class Space {
 
         public Space ()
--- a/Keas.Core/Domain/Space.cs
+++ b/Keas.Core/Domain/Space.cs
@@ -5,6 +5,7 @@ namespace Keas.Core.Domain
 {
     // Facilities link deptSpace, flattened
     // ties together rooms and departments
+    [Table("Spaces")]
     public class Space {
 
         public Space ()
--- a/Keas.Mvc/ClientApp/components/Spaces/SpacesContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Spaces/SpacesContainer.tsx
@@ -36,7 +36,7 @@ export default class SpacesContainer extends React.Component<{}, IState> {
     }
 
     public async componentDidMount() {
-        const spaces = await this.context.fetch(`/api/${this.context.team.slug}/spaces/list?teamSlug=${this.context.team.slug}`);
+        const spaces = await this.context.fetch(`/api/${this.context.team.slug}/spaces/list`);
         const tags = await this.context.fetch(`/api/${this.context.team.slug}/tags/listTags`);
         this.setState({ loading: false, spaces, tags });
     }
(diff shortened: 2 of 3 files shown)


jSylvestre:
@jcubbage Since you are changing the SpacesController, do you want to change the !String.Is... to !string.Is... (lower case s)

Keas.Mvc/Controllers/Api/SpacesController.cs
```diff
@@ -31,11 +31,12 @@ public async Task<IActionResult> SearchSpaces(string q)
             return Json(space);
         }
 
-        public async Task<IActionResult> List(string orgId)
+        public async Task<IActionResult> List(string teamSlug)
         {
             //TODO clean up workstations query
+            var orgIds = await _context.FISOrgs.Where(f => f.Team.Slug == teamSlug).Select(x => x.OrgCode).ToListAsync();
```
jSylvestre: distinct

jSylvestre:
`namespace Keas.Core.Domain
{
    // Facilities link deptSpace, flattened
    // ties together rooms and departments
    [Table("Space")]
    public class Space {
`


image

Keas.Mvc/ClientApp/components/Spaces/SpacesContainer.tsx
```diff
@@ -36,7 +36,7 @@ export default class SpacesContainer extends React.Component<{}, IState> {
     }
 
     public async componentDidMount() {
-        const spaces = await this.context.fetch(`/api/${this.context.team.slug}/spaces/list?orgId=ADNO`);
+        const spaces = await this.context.fetch(`/api/${this.context.team.slug}/spaces/list?teamSlug=${this.context.team.slug}`);
```
srkirkland: @jcubbage the slug is already in the URL why would you pass in twice or do I not want to know?

Keas.Mvc/Controllers/Api/SpacesController.cs
```diff
@@ -24,44 +23,44 @@ public async Task<IActionResult> SearchSpaces(string q)
...
-                || (!String.IsNullOrWhiteSpace(x.RoomNumber) && queryWords.Any(s => x.RoomNumber.ToLower().Contains(s))))
+                .Where(x => (!string.IsNullOrWhiteSpace(x.BldgName) && queryWords.Any(s => x.BldgName.ToLower().Contains(s)))
+                || (!string.IsNullOrWhiteSpace(x.RoomName) && queryWords.Any(s => x.RoomName.ToLower().Contains(s)))
+                || (!string.IsNullOrWhiteSpace(x.RoomNumber) && queryWords.Any(s => x.RoomNumber.ToLower().Contains(s))))
                 .AsNoTracking().ToListAsync();
             return Json(space);
         }
 
-        public async Task<IActionResult> List(string orgId)
+        public async Task<IActionResult> List(string teamSlug)
```
srkirkland: @jcubbage if you just take team name as a param here it should map properly, and then you can get rid of that teamSlug query string param you added

Keas.Core/Domain/Space.cs
```diff
@@ -5,6 +5,7 @@ namespace Keas.Core.Domain
 {
     // Facilities link deptSpace, flattened
     // ties together rooms and departments
+    [Table("Spaces")]
```
srkirkland: @jSylvestre real table is singular now right?

jSylvestre: @srkirkland Correct. Table with the data is "Space"

jcubbage: That's what I put!

PR #202, opened 2018-10-25
special route for admin pages

srkirkland:
admin was being interpreted as a regular team name, so we can either create a new route just for admin stuff, which i did, or we can update the "TeamRoutes" route and say that there are some "reserved" team names, like Admin.


jcubbage:
Should we ALSO limit team slugs to exclude controller names? Only SysAdmins can create teams, but that may be passed to others in the future.

PR #208, opened 2018-10-25
Jsc/active query filters

### LINKED ISSUE #184: make sure queries check Active flag

jcubbage:
closes #184. Alternative approach  to #187


jcubbage:
@srkirkland Yes: The reports have a flag to include inactive items. That flag is ignored if we don't ignore the query filter.

srkirkland:
@jcubbage ok 🚢

PR #212, opened 2018-10-26
Didn't actually want this after all.

jcubbage:


jSylvestre:
@jcubbage  Tests will fail. I'll fix

PR #214, opened 2018-10-26
validate email on edit person

### LINKED ISSUE #193: Edit Person
If you edit the email to something invalid (like two @ symbols) it doesn't save, but it doesn't tell the user that there was a problem either.

laholstege:
closes #193


laholstege:
@srkirkland i stole it from anlab so it will work as well as it does there

PR #225, opened 2018-10-29
Use a string extension and humanizer to do a title case for first and…

jSylvestre:
… last names.


jSylvestre:
FYI, same as with textInfo.TitleCase, it looks like this doesn't do all upper case to avoid changing acronyms

Issue #223, opened 2018-10-29
search attributes 


laholstege:
"search attributes" instead of "search attribute key"

PR #226, opened 2018-10-29
Jsc/wording updates

jcubbage:
Fixes issues #222 and #223


Keas.Mvc/ClientApp/components/Equipment/SearchAttributes.tsx
```diff
@@ -24,7 +24,7 @@ export default class SearchAttributes extends React.Component<IProps, {}> {
                     }}
                     selected={this.props.selected}
                     selectHintOnEnter={true}
-                    placeholder="Search for attributes"
+                    placeholder="Search attributes"
                     allowNew={true}
                     emptyLabel={''}
                     newSelectionPrefix="Select attribute key: "
```
srkirkland: @jcubbage @laholstege Since we aren't actually doing any new selection here and just always search, should we make the newSelectionPrefix prop just be "Search attributes" or "Search for: " or similar?

PR #229, opened 2018-10-30
Lh/fixes

laholstege:


Keas.Mvc/ClientApp/components/Spaces/SpacesTable.tsx
```diff
@@ -82,20 +82,29 @@ export default class SpacesTable extends React.Component<IProps, {}> {
...
                         ),
                     },
                     {
                         Header: "Workstations",
                         headerClassName: "table-10p",
                         className: "table-10p",
-                        accessor: "workstationsCount",
+                        id: "workstationsCount",
+                        accessor: spaceInfo => {
+                            return [spaceInfo.workstationsInUse,spaceInfo.workstationsTotal];
```
srkirkland: I think this would be a little more discoverable as an object instead of an array, since I find it hard to reason about what row.workstationsCount[1] is representing

Issue #227, opened 2018-10-29
asset search should pull all assets, not just assigned


laholstege:

laholstege:
#237 does this for equip -- need to add to keys

PR #231, opened 2018-10-30
Lh/equipment search

laholstege:
#227 
not really in the docs, but heres the option:
https://github.com/ericgio/react-bootstrap-typeahead/issues/86

image

### DIFF
--- a/Keas.Mvc/ClientApp/Types.ts
+++ b/Keas.Mvc/ClientApp/Types.ts
@@ -117,10 +117,6 @@ export interface IEquipmentAssignment {
   person: IPerson;
 }
 
-export interface IHasExpiration {
-  expiresAt: Date;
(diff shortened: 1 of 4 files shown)


laholstege:
@jcubbage @jSylvestre what do you think?

jcubbage:
That's not a bad idea....but I wonder if doing it the way you make it more confusing if they accidentally choose it. Like someone was just mousing over the list intending to go to the unassigned one, but click on the wrong one and all of a sudden they get a whole new screen with that equipment details. Also, how difficult would that to build since /equipment/1 does not currently open details on equip ID 1?

srkirkland:
@laholstege is going to change this so we order by disabled then by name, so we only show disabled after any possible enabled ones

PR #236, opened 2018-10-31
Default to the first tab (People)

jSylvestre:


jSylvestre:
FYI, has benefit that all admin roles can see people.

PR #251, opened 2018-11-07
update all of the packages that are easyish

srkirkland:
webpack still at v3, typescript at v2 still.


srkirkland:
@jSylvestre no rush but if you could run this locally and just make sure it works that would be awesome, just to have another computer try it out.  Make sure to run an install to get the new dependencies (I think VS might do that automatically for you, not sure)

jSylvestre:
@srkirkland 👍 But you currently have merge conflicts

PR #252, opened 2018-11-07
Jpk/keyserials

jpknoll:


Keas.Core/Models/ExpiringItemsEmailModel.cs
```diff
@@ -1,8 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
```
srkirkland: @jpknoll this makes the build fail for me

srkirkland:
@jpknoll just from running it, the main problem I see is when you are adding a serial from the person page.  It fails but I'm not sure what it should do even if it succeeds.  I think there is a missing step about what key to associate the serial with.

Also, I think most keys won't have names and it fails if you try to add a key w/o a name right now

Keas.Mvc/Program.cs
```diff
@@ -18,7 +12,14 @@ public class Program
         public static void Main(string[] args)
         {
             var host = BuildWebHost(args);
-            
+
+            using (var scope = host.Services.CreateScope())
+            {
+                var dbContext = scope.ServiceProvider.GetRequiredService<ApplicationDbContext>();
+                var dbInitilizer = new DbInitializer(dbContext);
+                dbInitilizer.Initialize();
```
srkirkland: @jpknoll danger, danger, danger

jpknoll: Go read the init function

srkirkland: not so dangerous apparently

PR #253, opened 2018-11-07
Lh/delete equipment and access

laholstege:
#145 
still need to add for keys


Keas.Mvc/ClientApp/components/Access/DeleteAccess.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    </ModalFooter>
+                </Modal>
+            </div>
+        );
+    }
+
+    private _deleteAccess = async () => {
+        if(this.props.selectedAccess.assignments.length > 0 &&
+            !confirm("This access is currently assigned, are you sure you want to delete it?")){
+            return false;
```
srkirkland: Probably no point to return false or true since this is essentially a void func.  Also do you think it would read cleaner if you inverted your `if`, something like `if (this.props.selectedAccess.assignments.length > 0 && confirm('...')) { await delete() }`

Keas.Mvc/Controllers/Api/AccessController.cs
```diff
@@ -130,15 +130,38 @@ public async Task<IActionResult> Delete([FromBody]Access access)
...
+            {
+                try
+                {
+                    _context.Access.Update(access);
+
+                    if(access.Assignments.Count > 0)
+                    {
+                        foreach(var assignment in access.Assignments.ToList()) 
+                        {
+                            await _eventService.TrackUnAssignAccess(assignment, Team); // call before we remove person info
```
srkirkland: Not a huge issue but under the covers this tracking does a saveChanges() which is probably resulting in just tons of DB calls that are unnecessary.  If it becomes a problem we should split out and let the caller handle the persistence

Keas.Mvc/Controllers/Api/AccessController.cs
```diff
@@ -130,15 +130,38 @@ public async Task<IActionResult> Delete([FromBody]Access access)
...
+                    access.Active = false;
+                    await _context.SaveChangesAsync();
+                    // TODO: track history?
+
+                    transaction.Commit();
+                    return Json(null);
+                }
+                catch(Exception)
+                {
+                    throw;
```
srkirkland: @laholstege if this is all you are doing there is no reason to do a try/catch block at all.  Just remove the whole thing and get one layer of nesting back

PR #272, opened 2018-11-14
Jsc/history update

### LINKED ISSUE #263: History should have item.name

jcubbage:
Fixes #263, and #262 
Note: No key delete in controller. No Workstation deletion in controller. No AccessAssignment Update in controller.


Keas.Mvc/Services/HistoryService.cs
```diff
@@ -442,6 +453,132 @@ public async Task<History> WorkstationAccepted(Workstation workstation)
             await _context.SaveChangesAsync();
             return historyEntry;
         }
+
+        public async Task<History> KeyDeleted(Key key)
+        {
+            var user = await _securityService.GetUser();
+            var historyEntry = new History
+            {
+                Description = key.Name != null ? "Key (" + key.Name + ") Deleted by " + user.Name : "Key (" + key.Code + ") Deleted by " + user.Name,
```
srkirkland: You might want to think about a utility method because this same code is repeated a tons of times and if we ever change the wording slightly it would be good to not have to change in so many places.

I think if you use one of the base classes like AssetBase you might be able to clean this up, say `GetDescription(asset, user, action)` or something where action is deleted in this case.  Each asset could have an override to `toString()` as well to be able to write it's own display so you can offload the work of null checking and such

jpknoll: Do not override toString()

Create an abstract GetAssetName method on the asset base and implement it with nameof(class)

srkirkland: sure that works

Keas.Mvc/Services/EventService.cs
```diff
@@ -23,6 +23,14 @@ public interface IEventService
         Task TrackAcceptKey(KeySerial keySerial);
         Task TrackAcceptEquipment(Equipment equipment);
         Task TrackAcceptWorkstation(Workstation workstation);
+        Task TrackKeyDeleted(Key key);
+        Task TrackAccessDeleted(Access access);
+        Task TrackEquipmentDeleted(Equipment equipment);
+        Task TrackWorkstationDeleted(Workstation workstation);
+        Task TrackAssignmentUpdatedKeySerial(KeySerial keySerial);
```
jpknoll: We should be consistently using either the parent asset (like here), or the assignment object (like below).

My vote is for the assignment object.

Keas.Core/Domain/AssetBase.cs
```diff
@@ -25,7 +25,14 @@ public AssetBase()
...
 
-        public bool Active { get; set; }    
+        public bool Active { get; set; }
+
+        public string GetDescription(string asset, string title, User user, string action)
+        {
+            return asset + "(" + title + ") " + action + " by " + user.Name;
+        }
+
+        public virtual string Title => Name;
```
jSylvestre: @jcubbage This was causing the tests to fail. I pushed a fix to the tests.

PR #273, opened 2018-11-14
Jpk/key serials 2

jpknoll:


Keas.Mvc/Services/HistoryService.cs
```diff
@@ -240,18 +240,17 @@ public async Task<History> EquipmentAssigned(Equipment equipment)
...
+        public async Task<History> KeySerialUnassigned(KeySerialAssignment keySerialAssignment)
         {
             var user = await _securityService.GetUser();
             var historyEntry = new History
             {
-                Description = "Key Unassigned  by " + user.Name,
+                Description = "Key Unassigned by " + user.Name,
                 ActorId = user.Id,
-                AssetType = "Key",
+                AssetType = "KeySerial",
```
jSylvestre: @jcubbage Can you confirm if the AssetType should be changed here? If it should be, @jpknoll You should probably change it on line 199 above as well.

Keas.Mvc/ClientApp/components/Keys/AssignKeySerial.tsx
```diff
@@ -46,18 +46,20 @@ export default class AssignKey extends React.Component<IProps, IState> {
   constructor(props: IProps) {
     super(props);
 
-    const date = (!!this.props.selectedKeySerial && !!this.props.selectedKeySerial.assignment) 
-      ? moment(this.props.selectedKeySerial.assignment.expiresAt)
+    const assignment = props.selectedKeySerial && props.selectedKeySerial.keySerialAssignment;
+
+    const date = (!!assignment)
```
srkirkland: what's the point of the double bang given we know from the line above this is boolean

jpknoll: It's not a bool.
The && is null coalesce.

srkirkland: still not seeing the point of it

jpknoll: `props.selectedKeySerial` can be null and the next several lines are a lot cleaner with the `assignment` instead of `props.selectedKeySerial.keySerialAssignment`.

PR #281, opened 2018-11-26
People are more than users

### LINKED ISSUE #275: Should we switch user in history service to person?

jcubbage:
Fixes #275  Use people instead of user. In our department, we have 2 Patrick Browns. So, we would need to differentiate them by name (Patrick J Brown and Patrick M Brown), then use the person, not user name to get them listed correctly in history.

### DIFF
--- a/Keas.Core/Domain/AssetBase.cs
+++ b/Keas.Core/Domain/AssetBase.cs
@@ -29,7 +29,7 @@ namespace Keas.Core.Domain
 
         public string GetDescription(string asset, string title, Person person, string action)
         {
-            return asset + "(" + title + ") " + action + " by " + person.Name;
+            return asset + "(" + title + ") " + action + " by " + person.Name + " (" + person.UserId + ")";
         }
 
         public virtual string Title => Name;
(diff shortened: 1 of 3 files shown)


jpknoll:
Not totally sold on this being the solution to non-repudiation. Won't they both have different user ids?

jcubbage:
Yes, but we don't display the userID in the text of the history description. Database will always know they are different, but folks reading the text won't.

jpknoll:
What's to prevent Person.Name being the same for two people?

jcubbage:
We have an open issue to allow teams to edit the person's name (only for that team). Haven't implemented, but that's how teams will be able to differentiate.

jpknoll:
Yeah, but shouldn't the user be capable of editing their own User.DisplayName? You can update the history descriptions to be "DisplayName (kerberos)" to differentiate, which will show the user's _preferred_ display name but still offer the feature you need.

jcubbage:
In the design committee meetings, they were very insistent that team admins be able to update the name for that team, but no change that person's name in another team. That's why we created the person.firstname and .lastname instead of just using the user name that is pulled from IAM. That's why I updated to pull from person instead of user.

jpknoll:
Okay. Given that requirement, yeah, use Person. But I still think you should a "(kerberos)" entry instead of mandating Team Unique.

PR #286, opened 2018-11-27
Jsc/history person

jcubbage:
Check to see if references passed in. If not, create needed ones.

### DIFF
--- a/Keas.Mvc/Services/HistoryService.cs
+++ b/Keas.Mvc/Services/HistoryService.cs
@@ -2,6 +2,7 @@
 using Keas.Core.Data;
 using Keas.Core.Domain;
 using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
 
 namespace Keas.Mvc.Services
 {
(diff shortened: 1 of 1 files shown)


Keas.Mvc/Services/HistoryService.cs
```diff
@@ -58,8 +59,9 @@ public HistoryService(ApplicationDbContext context, ISecurityService securitySer
         }
 
         public async Task<History> KeyCreated(Key key)
-        {            
-            var person = await _securityService.GetPerson(key.Team.Slug);
+        {
+            var team = await _context.Teams.Where(t => t.Id == key.TeamId).Select(t => t.Slug).SingleAsync();
+            var person = await _securityService.GetPerson(team);
```
srkirkland: @jcubbage these two lines should be in a helper, maybe in the security service, which can get person by TeamId.

jcubbage: Better?

PR #294, opened 2018-11-28
Lh/edit and delete person

### LINKED ISSUE #111: person edit info
supervisor (future)
phone
title
notes?

laholstege:
closes #111 
may want to do more validation? only validates that name exists and that email exists and is valid
also probably need @cydoval to fix up styles/what info is shown on details page

also: 
* adds deleting of a person
* adds confirm on delete for all types 
* adds link to people/create for `AssignPerson`

### DIFF
--- a/Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
@@ -68,7 +68,7 @@ export default class PeopleContainer extends React.Component<{}, IState> {
             this._renderTableView(personAction === "create")
           }
           {personAction === "details" && !!detailPerson && !!detailPerson.person &&
-            this._renderDetailsView(detailPerson.person)
+            this._renderDetailsView(detailPerson)
           }
         </div>
       </div>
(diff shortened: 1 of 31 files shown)


Keas.Mvc/ClientApp/components/People/PeopleContainer.tsx
```diff
@@ -208,6 +209,36 @@ export default class PeopleContainer extends React.Component<{}, IState> {
...
+  const deleted: IPerson = await this.context.fetch(`/api/${this.context.team.slug}/people/delete`, {
+    body: JSON.stringify(person),
+    method: "POST"
+  });
+
+  this._goBack();
+
+  // update already existing entry in key
+  const updatePeople = [...this.state.people];
+  updatePeople.splice(index, 1);
```
srkirkland: this is a cool way of doing it

Keas.Mvc/ClientApp/components/People/PersonDetails.tsx
```diff
@@ -2,34 +2,32 @@ import * as PropTypes from 'prop-types';
...
     onEdit: (person: IPerson) => void;
+    onDelete: (person: IPerson) => void;
 }
 
 
 export default class PersonDetails extends React.Component<IProps, {}> {
 
     public render() {
-        if (this.props.selectedPerson == null)
+        if (this.props.selectedPersonInfo == null || this.props.selectedPersonInfo.person == null)
```
srkirkland: @laholstege I know it was like that before but try to dump the double equals if you can.  Either use `===` or cleaner yet just use `!` because `if (!this.props.selectedPersonInfo)` should get you the same result

jSylvestre:
@laholstege 🚢 DB in prod has been updated.

Issue #221, opened 2018-10-29
other spaces details


laholstege:
square footage
~~tags~~
room use (not roomCategoryName)

jSylvestre:
@laholstege What was this? Or did you mean to assign @jcubbage ?

jSylvestre:
![image](https://user-images.githubusercontent.com/469010/49253521-b54e6000-f3db-11e8-9c43-8f0711ddc9a6.png)

PR #307, opened 2018-11-29
More Spaces details

jSylvestre:
Issue #221 

image


jSylvestre:
maybe we don't want all these?

image

PR #317, opened 2018-12-10
Lh/key fixes

### LINKED ISSUE #313: create key just goes to blank screen
the key gets created but the routing doesn't work and maybe other stuff too.  The screen just goes blank.

laholstege:
closes #313

### DIFF
--- a/Keas.Mvc/ClientApp/components/Keys/KeyTable.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/KeyTable.tsx
@@ -160,26 +160,6 @@ export default class KeyTable extends React.Component<IProps, IState> {
         );
     }
 
-    private renderAvailableColumn = (row: IRow) => {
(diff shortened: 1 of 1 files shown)


Keas.Mvc/ClientApp/components/Keys/KeyTable.tsx
```diff
@@ -81,13 +81,61 @@ export default class KeyTable extends React.Component<IProps, IState> {
                                 .includes(filter.value.toLowerCase())
                     },
                     {
-                        Header: "",
-                        headerClassName: "table-actions",
-                        filterable: false,
-                        sortable: false,
-                        resizable: false,
-                        className: "table-actions",
-                        Cell: this.renderAvailableColumn
```
jpknoll: You've both fixed the render method below, and replaced with an inline method. Pick one.

PR #339, opened 2019-01-07
Jpk/logging tec

jpknoll:


Keas.Core/Services/IEmailService.cs
```diff
@@ -70,200 +51,145 @@ public async Task SendExpiringMessage(int personId, ExpiringItemsEmailModel mode
...
+                .Select(u => new Recipient() {Address = new Address(u.Email, u.Name, "cc")})
+                .ToList();
+
+#if !DEBUG
+            // add emails to
+            ccEmails.ForEach(transmission.Recipients.Add);
+#endif
+
+            // build view
+            var engine = GetRazorEngine();
```
jSylvestre: @jpknoll Did you test this? We had to jump through some hoops using RazorLightEngineBuilder.

Keas.Core/Services/IEmailService.cs
```diff
@@ -70,200 +51,145 @@ public async Task SendExpiringMessage(int personId, ExpiringItemsEmailModel mode
...
-                }
-                else if (access.ExpiresAt > DateTime.UtcNow.AddDays(1))
-                {
-                    access.NextNotificationDate = access.ExpiresAt.AddDays(-1);
-                }
-                else
-                {
-                    access.NextNotificationDate = DateTime.UtcNow.AddDays(1);
-                }
-                _dbContext.AccessAssignments.Update(access);
```
jSylvestre: @jpknoll Was the update statement not needed?

Keas.Core/Services/IEmailService.cs
```diff
@@ -272,20 +198,35 @@ public async Task SendNotificationMessage(User user)
                 _dbContext.Notifications.Update(notification);                
             }
             await _dbContext.SaveChangesAsync();
+        }
 
-            var mimeType = new System.Net.Mime.ContentType("text/html");
-
-            var alternate = AlternateView.CreateAlternateViewFromString(message.Body, mimeType);
-            message.AlternateViews.Add(alternate);
```
jSylvestre: @jpknoll Does Sparkpost have AlternateViews?

Keas.Core/Services/IEmailService.cs
jpknoll: No, the objects are already in tracking. Just need the save at the end.

Keas.Core/Services/IEmailService.cs
jpknoll: You can specify .Text and .Html views on the Transaction.Content.

Keas.Core/Services/IEmailService.cs
jSylvestre: @jpknoll Needed, probably because the initial query has .AsNoTracking()

jpknoll: Then we should remove the AsNoTracking, because we want to track them :-D

jSylvestre: @jpknoll I think that caused a lot of other problems. But maybe @jcubbage knows for sure

Issue #341, opened 2019-01-08
Line Width


jSylvestre:
https://peaks.ucdavis.edu/caesdo/Report

![image](https://user-images.githubusercontent.com/469010/50846794-dbddb000-1324-11e9-8bfe-b7a39a3f5f3d.png)

PR #343, opened 2019-01-08
Report Page uses card-header-primary...

jSylvestre:
The other cards have a thicker line

Old:
image

New:
image


jSylvestre:
Closes #341

PR #358, opened 2019-01-11
Supervisor report

### LINKED ISSUE #352: Report
List all people a supervisor supervises

jcubbage:
Closes #352 
Do we need to display anything else? I left it as a table in case we want email/phone/etc. Otherwise we should swap out to a bulleted list (maybe ordered list?).


jSylvestre:
@jcubbage 
1) You need a link to the report
2) Maybe provide more columns, about the user and a link to them on the asset page?

image

PR #367, opened 2019-01-15
Missing includes

### LINKED ISSUE #366: Edit equipment to change exp date
Throws exception because the person is null:

image

This is in the Assignment.GetDescription() method. (Note, passed person is ok,pberon in the object isn't. Probably needs an include?

jSylvestre:
needed for when exp date is updated.
Close #366


jSylvestre:
The other assignments seemed ok.

PR #380, opened 2019-01-17
Lh/details on list

laholstege:
image


Keas.Mvc/ClientApp/components/Access/AccessList.tsx
```diff
@@ -32,6 +32,7 @@ export default class AccessList extends React.Component<IProps, {}> {
       <table className="table">
         <thead>
           <tr>
+            <th/>
```
srkirkland: @laholstege this might work but I don't think the `th` tag is self-closing.  I think valid HTML would require it to be `<th></th>`

laholstege: @srkirkland but if i isnt this way it angers the linter "JSX elements with no children must be self-closing"

PR #398, opened 2019-01-17
Jsc/team api

jcubbage:
How's this look? Needed to add new column, so tests are failing.

### DIFF
--- a/Keas.Mvc/Controllers/FeedController.cs
+++ b/Keas.Mvc/Controllers/FeedController.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Keas.Core.Data;
+using Keas.Mvc.Models;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using Keas.Core.Domain;
+
+namespace Keas.Mvc.Controllers
+{
+    public class FeedController : SuperController
+    {
+         private readonly ApplicationDbContext _context;
+
+        public FeedController(ApplicationDbContext context)
+        {
+            this._context = context;
+        }
+        public async Task<IActionResult> TeamFeed(Guid id, string space)
+        {            
+            if(space == "yes")
+            {
+                var people = await _context.People
+                .Where(x=> x.Team.Slug == Team && x.Active && x.Team.ApiCode == id)
+                .Select(p => new {p.Name, p.FirstName, p.LastName, p.Email, p.UserId, p.Title, p.TeamPhone, p.Tags,
+                workstation = (from w in _context.Workstations where w.Assignment.PersonId == p.Id select w)
+                    .Include(w => w.Space)
+                    .Select(w => new {w.Name, w.Space.BldgName, w.Space.RoomNumber})
+                })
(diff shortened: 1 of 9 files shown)


Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+         private readonly ApplicationDbContext _context;
+
+        public FeedController(ApplicationDbContext context)
+        {
+            this._context = context;
+        }
+        public async Task<IActionResult> TeamFeed(Guid id, string space)
+        {            
+            if(space == "yes")
```
srkirkland: @jcubbage break these two paths into different methods so everything isn't in the one method

Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class FeedController : SuperController
+    {
+         private readonly ApplicationDbContext _context;
+
+        public FeedController(ApplicationDbContext context)
+        {
+            this._context = context;
+        }
+        public async Task<IActionResult> TeamFeed(Guid id, string space)
```
srkirkland: Name it something like `includeSpace` so we know it's a bool

Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -0,0 +1,45 @@
...
+        {            
+            if(space == "yes")
+            {
+                var people = await _context.People
+                .Where(x=> x.Team.Slug == Team && x.Active && x.Team.ApiCode == id)
```
srkirkland: Instead of check the api code by adding it in here, you should do its own check first before going to the DB.  That way you can return the proper unauthorized status, instead of the 200 success status with no results they will get with a bad key now

Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -0,0 +1,45 @@
...
+                .Select(p => new {p.Name, p.FirstName, p.LastName, p.Email, p.UserId, p.Title, p.TeamPhone, p.Tags,
+                workstation = (from w in _context.Workstations where w.Assignment.PersonId == p.Id select w)
```
srkirkland: Should this be `Workstations` plural, and capitalized, to match the other fields?

Keas.Mvc/Controllers/FeedController.cs
jcubbage: I origanally had that, but the issue wanted it as one with a parameter :)

srkirkland: @jcubbage The one parameter is fine and can stay the same, I just think you should make a private method or two to handle the actual logic of it.

Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -0,0 +1,67 @@
...
+        public FeedController(ApplicationDbContext context)
+        {
+            this._context = context;
+        }
+        public async Task<IActionResult> TeamFeed(Guid id, string includeSpace)
+        {            
+             var validKey = await _context.Teams.Where(t => t.Slug == Team && t.ApiCode == id).AnyAsync();
+
+            if(!validKey){
+                Unauthorized();
```
srkirkland: Need to `return Unauthorized()`

Issue #406, opened 2019-01-17
tooltip for showing that 1/2 on table means available/total 


laholstege:

jcubbage:
Added tooltip #414

PR #414, opened 2019-01-22
Added tooltip

jcubbage:
Closes Issue #406

### DIFF
--- a/Keas.Mvc/ClientApp/components/Keys/KeyList.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/KeyList.tsx
@@ -25,7 +25,7 @@ export default class KeyList extends React.Component<IProps, {}> {
                         <th/>
                         <th>Name</th>
                         <th>Code</th>
-                        <th className="">Available Serials</th>
+                        <th className="" title="In Use / Total">Available Serials<sup><i class="fas fa-info-circle"></i></sup></th>
                         <th className="list-actions">Actions</th>
                     </tr>
                 </thead>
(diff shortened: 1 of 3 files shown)


Keas.Mvc/ClientApp/components/Keys/KeyList.tsx
```diff
@@ -25,7 +26,7 @@ export default class KeyList extends React.Component<IProps, {}> {
                         <th/>
                         <th>Name</th>
                         <th>Code</th>
-                        <th className="">Available Serials</th>
+                        <th className=""><span id="serialTooltip">Available Serials<sup><i class="fas fa-info-circle"></i></sup></span><UncontrolledTooltip placement="bottom" target="serialTooltip">In Use / Total</UncontrolledTooltip></th>
```
srkirkland: @jcubbage probably a little picky here but `sup` shouldn't be used for presentation, just for like exponents and such. Instead you'd use a `vertical-align: super` css property.  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup

srkirkland:
Also obviously something is wrong with the webpack build so one of the react html elements you added is causing problems.

Keas.Mvc/ClientApp/components/Keys/KeyList.tsx
laholstege: @jcubbage i would also probably move it so the `span` only applies to the icon, and have it open to the right instead of the bottom 
also, might as well remove `className=""`

PR #439, opened 2019-01-28
Jsc/pend accept report

### LINKED ISSUE #428: Unaccepted Items Report(s)
Similar to the expiring items report(s) I guess.

jcubbage:
One report limits to only the roles you have. Should we update expiry to be the same view?
Closes #428

### DIFF
--- a/Keas.Mvc/Models/ExpiringItemsViewModel.cs
+++ b/Keas.Mvc/Models/ExpiringItemsViewModel.cs
@@ -10,7 +10,7 @@ using Microsoft.Extensions.Options;
 
 namespace Keas.Mvc.Models
 {
-    public class ExpiringItemsViewModel
+    public class ReportItemsViewModel
     {
         public AccessAssignment[] Access { get; set; }
         public KeySerial[] Keys { get; set; }
@@ -29,7 +29,7 @@ namespace Keas.Mvc.Models
         public String ShowType { get; set; }
        
 
-        public static async Task<ExpiringItemsViewModel> Create(ApplicationDbContext context, DateTime expiresBefore, string teamName, bool showInactive, string showType)
+        public static async Task<ReportItemsViewModel> Create(ApplicationDbContext context, DateTime expiresBefore, string teamName, bool showInactive, string showType)
         {
             
             var expiringAccess = await context.AccessAssignments.IgnoreQueryFilters().Where(a => (showType == "All" || showType == "Access") &&
@@ -46,7 +46,7 @@ namespace Keas.Mvc.Models
                 .Include(w => w.Assignment).ThenInclude(a=> a.Person).AsNoTracking().ToArrayAsync();
 
             var itemList = new List<string>(new string[] {"All", "Access", "Equipment", "Key", "Workstation"});
-            var viewModel = new ExpiringItemsViewModel
+            var viewModel = new ReportItemsViewModel
             {
                 Access = expiringAccess,
                 Keys = expiringKey,
--- a/Keas.Mvc/Models/ExpiringItemsViewModel.cs
+++ b/Keas.Mvc/Models/ExpiringItemsViewModel.cs
@@ -62,26 +62,26 @@ namespace Keas.Mvc.Models
             return viewModel;
         }
 
-        public static async Task<ReportItemsViewModel> CreateUnaccepted(ApplicationDbContext context, string teamName, bool showInactive, string showType, List<Role> userRoles)
+        public static async Task<ReportItemsViewModel> CreateUnaccepted(ApplicationDbContext context, string teamName, bool showInactive, string showType, List<Role> userRoles, ISecurityService _securityService)
         {  
             var expiringAccess = await context.AccessAssignments.IgnoreQueryFilters().Where(a => (showType == "All" || showType == "Access") && 
-                (userRoles.Where(r => r.Name == Role.Codes.DepartmentalAdmin).Any() || userRoles.Where(r => r.Name == Role.Codes.AccessMaster).Any() || userRoles.Where(r => r.Name == Role.Codes.Admin).Any()) &&
+                (_securityService.IsRoleOrDAInList(userRoles, Role.Codes.AccessMaster)) &&
                 a.Access.Team.Slug == teamName && !a.IsConfirmed && (a.Access.Active || a.Access.Active == !showInactive))
                 .Include(a => a.Access).Include(a=> a.Person).AsNoTracking().ToArrayAsync();
             var expiringKey = await context.KeySerials.IgnoreQueryFilters().Where(a => (showType == "All" || showType == "Key") &&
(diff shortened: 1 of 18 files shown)


Keas.Mvc/Models/ExpiringItemsViewModel.cs
```diff
@@ -55,9 +57,70 @@ public static async Task<ExpiringItemsViewModel> Create(ApplicationDbContext con
...
                 ItemList =  itemList,
-                ShowType = showType,                
+                ShowType = showType               
             };
             return viewModel;
         }
+
+        public static async Task<ReportItemsViewModel> CreateUnaccepted(ApplicationDbContext context, string teamName, bool showInactive, string showType, List<Role> userRoles)
+        {  
+            var expiringAccess = await context.AccessAssignments.IgnoreQueryFilters().Where(a => (showType == "All" || showType == "Access") &&
```
srkirkland: I don't really like the idea here of adding queries to the iqueryable that don't have anything to do with the underlying db.  How does this get translated to SQL?  Are they just ignored or does it break up the SQL in some way?  I think overall if you just threw in a little private method which would determine if the info should show for a given type (i.e. "Workstation") it might simplify matters a lot and not be so repetitive. Also I'm wondering what happens from a SQL level if you don't have access to something?  Does the query short circuit before the SQL is generated based off the local arguments like the showType == "All"?

Keas.Mvc/Models/ExpiringItemsViewModel.cs
```diff
@@ -55,9 +57,70 @@ public static async Task<ExpiringItemsViewModel> Create(ApplicationDbContext con
...
+                ShowType = showType                
+            };
+            return viewModel;
+        }
+
+        public static List<string> populateItemList(List<Role> userRoles) 
+        {
+            var itemList = new List<string>() {"All"};
+
+            if(userRoles.Where(r => r.Name == Role.Codes.DepartmentalAdmin).Any() || userRoles.Where(r => r.Name == Role.Codes.AccessMaster).Any() || userRoles.Where(r => r.Name == Role.Codes.Admin).Any())
```
srkirkland: Maybe throw this into the security service?  This is at least 8 times in this one file where the exact same check is made.  It'll be hell to change if we ever add a new role

Keas.Mvc/Services/SecurityService.cs
```diff
@@ -244,6 +246,26 @@ public async Task<List<Role>> GetUserRolesInTeamOrAdmin(Team team)
...
+        {
+            var userId = _contextAccessor.HttpContext.User.Identity.Name;
+
+            var userPermissions = await _dbContext.TeamPermissions
+                .Where(x => x.Team.Slug == teamSlug && x.User.Id == userId)
+                .Select(tp => tp.Role)
+                .AsNoTracking()
+                .ToListAsync();
+
+            var admin = await _dbContext.SystemPermissions
```
srkirkland: @jcubbage This is good, but also we should discuss if Admins should be able to see this info or not.  I would argue a system admin should not be able to see a team's expiring keys, for example.  An admin should be able to emulate, create teams, and probably add themselves to someone else's team.  But I think it would simplify so much if we stopped checking for admin permissions

jcubbage: That makes sense....but I'd say make a new issue and do that in a new branch since so much of the app will get updated. Maybe make only the TeamAdmin page include Admins (so they can ass uses to the team). That way, if you want to get to assets or reports, you need to emulate. Restrict the policy to be team permissions only, and decorate only desired actions with admin.

Keas.Mvc/Models/ExpiringItemsViewModel.cs
jcubbage: Ignored:   The LINQ expression 'where ([r].Name == "DepartmentalAdmin")' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'where ([r].Name == "DepartmentalAdmin")' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'Any()' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'Any()' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'where ([r].Name == "AccessMaster")' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'where ([r].Name == "AccessMaster")' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'Any()' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'Any()' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'where ([r].Name == "Admin")' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'where ([r].Name == "Admin")' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'Any()' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'Any()' could not be translated and will be evaluated locally.
warn: Microsoft.EntityFrameworkCore.Query[20500]
      The LINQ expression 'where (((True AndAlso (({from Role r in __userRoles_0 where ([r].Name == "DepartmentalAdmin") select [r] => Any()} OrElse {from Role r in __userRoles_1 where ([r].Name == "AccessMaster") select [r] => Any()}) OrElse {from Role r in __userRoles_2 where ([r].Name == "Admin") select [r] => Any()})) AndAlso ([a.Access.Team].Slug == __teamName_3)) AndAlso Not([a].IsConfirmed))' could not be translated and will be evaluated locally.
Microsoft.EntityFrameworkCore.Query:Warning: The LINQ expression 'where (((True AndAlso (({from Role r in __userRoles_0 where ([r].Name == "DepartmentalAdmin") select [r] => Any()} OrElse {from Role r in __userRoles_1 where ([r].Name == "AccessMaster") select [r] => Any()}) OrElse {from Role r in __userRoles_2 where ([r].Name == "Admin") select [r] => Any()})) AndAlso ([a.Access.Team].Slug == __teamName_3)) AndAlso Not([a].IsConfirmed))' could not be translated and will be evaluated locally.

jcubbage: I'll experiment and see if the query fires off SQL even if the other bits are false.

Keas.Mvc/Models/ExpiringItemsViewModel.cs
jcubbage: :thumbsup:

Keas.Mvc/Services/SecurityService.cs
srkirkland: Yea I agree, I'll make a new issue for later

PR #469, opened 2019-02-01
Jsc/csv import

### LINKED ISSUE #355: Bulk load of keys
From csv file

jcubbage:
Closes #355 
First pass. Needs work for sure. Probably pull most of the code out of controller into service. But, I wanted a initial thoughts.


Keas.Mvc/Controllers/TeamAdminController.cs
```diff
@@ -363,5 +366,102 @@ public async Task<IActionResult> RemovePpsDepartment(int id, TeamPpsDepartment p
...
+                Message = "File not selected";
+                return RedirectToAction("Upload");
+            }
+            // Add counts
+            var team = await _context.Teams.FirstAsync(t => t.Slug == Team);
+            using (var reader = new StreamReader(file.OpenReadStream()))
+            using (var csv = new CsvReader(reader))
+            {    
+               csv.Configuration.PrepareHeaderForMatch = (string header, int index) => header.ToLower();
+                 var record = new KeyImport();
```
jSylvestre: This indent...

Keas.Mvc/Controllers/TeamAdminController.cs
```diff
@@ -363,5 +369,172 @@ public async Task<IActionResult> RemovePpsDepartment(int id, TeamPpsDepartment p
...
+                                serial = new KeySerial();
+                                serial.Number = r.SerialNumber;
+                                serial.Name = r.SerialNumber;
+                                serial.Key = key;
+                                serial.Status = "Active";
+                                if(r.Status == "L" || r.Status == "Lost")
+                                {
+                                    serial.Status = "Lost";
+                                }
+                                if(r.Status == "B" || r.Status == "Destroyed")
```
jSylvestre: @jcubbage We shouldn't be translating "their" codes. They should replace "B" in their file.

Keas.Mvc/Models/KeyImport.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    [Required]
+    public string Keynumber { get; set; }
+    [Required]
+    public string  SerialNumber { get; set; }
+    public string Description { get; set; }
+
+    public string KerbUser { get; set; }
+
+    public DateTime DateIssued { get; set; }
```
jSylvestre: @jcubbage Needs to be nullable and you need to validate when you read it. (HasValue, isn't in the future, dueDate isn't in the past, etc.)

Keas.Mvc/Services/IdentityService.cs
```diff
@@ -22,19 +22,82 @@ public interface IIdentityService
...
         private readonly ApplicationDbContext _context;
+    
 
         public IdentityService(IOptions<AuthSettings> authSettings, ApplicationDbContext context)
         {
             _authSettings = authSettings.Value;
             _context = context;
         }
 
+        public async Task<(Person Person, int peopleCount)> GetOrCreatePersonFromKerberos(string kerb, int teamId)
```
srkirkland: Tuples?  Tuples!

jSylvestre:
@jcubbage KeySerialAssignments. You are populating ApprovedAt (I don't think we are using this, but it is fine)

Keys.Tags is NULL instead of an empty string. (Might be a problem)

jSylvestre:
@srkirkland This failed on appveyor because "don't" in the commit message

PR #477, opened 2019-02-06
Jcs/prevent duplicate key code

### LINKED ISSUE #473: Duplicate Keys
Prevent a key, and maybe a key Serial from being created that would be a duplicate.

Ignore inactive as there is no way to "reactivate" a key.

jSylvestre:
close #473


jcubbage:
I agree with the concept....but I'm assuming we can throw an error message better so react can detect and display a message?

PR #481, opened 2019-02-08
Jsc/supervisor link

### LINKED ISSUE #480: Supervisor Report
Have a link from a supervisor person details page to the supervisor's report.

jcubbage:
Closes #480

### DIFF
--- a/Keas.Mvc/Resources/PeopleQueries.cs
+++ b/Keas.Mvc/Resources/PeopleQueries.cs
@@ -1,6 +1,7 @@
 public static class PeopleQueries
 {
-    public static string List = @"select People.*, SUP.FirstName SupervisorFirstName, SUP.LastName SupervisorLastName, SUP.Email SupervisorEmail, SUP.UserId SupervisorUserId, EquipmentCount, AccessCount, KeyCount, WorkstationCount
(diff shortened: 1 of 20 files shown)


jSylvestre:
image

Keas.Mvc/Resources/PeopleQueries.cs
```diff
@@ -1,6 +1,7 @@
 public static class PeopleQueries
 {
-    public static string List = @"select People.*, SUP.FirstName SupervisorFirstName, SUP.LastName SupervisorLastName, SUP.Email SupervisorEmail, SUP.UserId SupervisorUserId, EquipmentCount, AccessCount, KeyCount, WorkstationCount
+    public static string List = @"select People.*, SUP.FirstName SupervisorFirstName, SUP.LastName SupervisorLastName, SUP.Email SupervisorEmail, SUP.UserId SupervisorUserId, EquipmentCount, AccessCount, KeyCount, WorkstationCount,
+    cast(CASE WHEN EXISTS(SELECT * FROM People emp WHERE emp.SupervisorId = people.id) THEN 1 ELSE 0 END as bit) as isSupervisor
```
srkirkland: @jcubbage maybe here do SELECT ID instead of select *.  It's possible SQL Server is smart enough to know that it is only needed for existence, but I wouldn't count on it and would rather not select out a ton of info when not used.

PR #494, opened 2019-02-13
Jsc/space key display

jcubbage:
Updates UI wording and limits counts to only active keys.

### DIFF
--- a/Keas.Mvc/Controllers/Api/KeysController.cs
+++ b/Keas.Mvc/Controllers/Api/KeysController.cs
@@ -28,34 +28,34 @@ namespace Keas.Mvc.Controllers.Api
         {
             var comparison = StringComparison.InvariantCultureIgnoreCase;
 
-            var keys = 
+            var keys =
             from key in _context.Keys
-                .Where(x => x.Team.Slug == Team 
+                .Where(x => x.Team.Slug == Team
                         && x.Active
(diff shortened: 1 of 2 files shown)


Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -42,10 +42,9 @@ select new
                 key = key,
                 id = key.Id,
                 code = key.Code,
-                serialsTotalCount = 
-                    (key.Serials).Count(),
+                serialsTotalCount = (from s in key.Serials where s.Status == "Active" select s).Count(),
                 serialsInUseCount = 
-                    (from s in key.Serials where s.KeySerialAssignment != null select s ).Count(),
+                    (from s in key.Serials where s.KeySerialAssignment != null && s.Status == "Active" select s ).Count(),
```
srkirkland: @jcubbage i'm not sure this one is worth doing.  We shouldn't have a non-active serial assigned to someone.  And if we do, it should be in-use.  I know right now we do allow that... but I'd say remove this and we'll fix the problem on the other end?  Alternately we can keep this in to be safe, though I think it's just extra db work we don't need

jSylvestre: I'd say keep this way until the other way is fixed.

PR #502, opened 2019-02-15
Lh/keys

### LINKED ISSUE #310: Assign Key Serial
A nice error is not displayed when a duplicate KeySerial is being assigned to a different person:

image

### LINKED ISSUE #399: key serial "create new"
add link "go to key management" since we don't have "create new key/serial" link

### LINKED ISSUE #468: have key serials show already assigned / non-active greyed out in list when searching

laholstege:
* serials now have similar search/create pattern as the rest of the assets. shows already assigned ones as greyed out w/ link to details.
if you're on key/details, the search only searches the key serials for that key
closes #468 closes #310 
image

* can now create a new serial from people/details, you just have to select a key first
closes #399 

* code is now a unique identifier for a key, and you can't add duplicates
image

### DIFF
--- a/Keas.Core/Domain/Key.cs
+++ b/Keas.Core/Domain/Key.cs
@@ -19,7 +19,6 @@ namespace Keas.Core.Domain
         protected internal static void OnModelCreating(ModelBuilder builder)
         {
             builder.Entity<Key>().HasQueryFilter(a => a.Active);
-            builder.Entity<Key>().HasIndex(x => x.Code).IsUnique();
         }
     }
 }
--- a/Keas.Core/Domain/Key.cs
+++ b/Keas.Core/Domain/Key.cs
@@ -19,6 +19,7 @@ namespace Keas.Core.Domain
         protected internal static void OnModelCreating(ModelBuilder builder)
         {
             builder.Entity<Key>().HasQueryFilter(a => a.Active);
+            builder.Entity<Key>().HasIndex(x => x.Code).IsUnique();
         }
     }
 }
--- a/Keas.Mvc/ClientApp/components/Keys/CreateKey.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/CreateKey.tsx
@@ -11,6 +11,7 @@ interface IProps {
     onOpenModal: () => void;
     closeModal: () => void;
     searchableTags: string[];
+    keyCodes: string[];
 }
 
 interface IState {
@@ -86,6 +87,7 @@ export default class CreateKey extends React.Component<IProps, IState> {
                             searchableTags={searchableTags}
                         />
                     </form>
+                    {this.state.error}
                 </ModalBody>
                 <ModalFooter>
                     <Button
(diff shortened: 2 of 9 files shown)


Keas.Core/Domain/Key.cs
```diff
@@ -19,6 +19,7 @@ public class Key : AssetBase
         protected internal static void OnModelCreating(ModelBuilder builder)
         {
             builder.Entity<Key>().HasQueryFilter(a => a.Active);
+            builder.Entity<Key>().HasIndex(x => x.Code).IsUnique();
```
laholstege: @srkirkland @jSylvestre does this count as a db change?

jcubbage: If we do this, and someone deletes a key, does that mean they can not recreate it if the deletion needs to be reversed? We don't currently have a way to move !active => active. Current accepted method was go create a new one.

jSylvestre: @laholstege Yes, I think so. And no, we can't do it. (Unique within team and when active)

srkirkland: @laholstege yea i think @jcubbage is right and we can't do this as a unique index.  It would mean that a given code could only be used by one team, ever, and they wouldn't be able to inactivate it.   Instead you should get rid of this and just do a SQL check in code to make sure it's unique among active keys.

Keas.Mvc/ClientApp/components/Keys/CreateKey.tsx
```diff
@@ -155,10 +157,13 @@ export default class CreateKey extends React.Component<IProps, IState> {
             valid = false;
         } else if (!key.code) {
             valid = false;
-            error = "You must give this key a name.";
+            error = "You must give this key a code.";
         } else if (key.code.length > 64) {
             valid = false;
-            error = "The name you have chosen is too long";
+            error = "The code you have chosen is too long";
+        } else if (this.props.keyCodes.indexOf(key.code) !== -1) {
```
srkirkland: i'm thinking this will do what you want to make sure they aren't already using that key code actively.  An extra DB check would be nice but i don't think it's that necessary as the downside is small

Keas.Mvc/ClientApp/components/Keys/CreateKey.tsx
```diff
@@ -11,6 +11,7 @@ interface IProps {
     onOpenModal: () => void;
     closeModal: () => void;
     searchableTags: string[];
+    keyCodes: string[];
```
srkirkland: I kinda think I'd prefer instead of passing along the entire list of keycodes, we instead pass a function which can be called to determine if a keycode already exists.  unless there is another reason createKey should know the full list of codes

Keas.Mvc/ClientApp/components/Keys/KeyContainer.tsx
```diff
@@ -267,6 +266,11 @@ export default class KeyContainer extends React.Component<IProps, IState> {
         return keyInfo;
     };
 
+    private _checkIfKeyCodeIsValid = (code: string) => {
+        const keyCodes = this.state.keys.map(x => x.key.code);
+        return keyCodes.indexOf(code) === -1;
```
srkirkland: @laholstege this works fine, but you could do a one-liner w/ either .find or .some or .findIndex.  I think I'd do it something like `return this.state.keys.some(x => x.key.code === code)`

PR #504, opened 2019-02-16
My googlefoo was strong on this one.

### LINKED ISSUE #467: tags on spaces details are messed up
image

jcubbage:
Closes #467 
Not sure if there is a better way....but I make an array from CSV, reduce it, then create CSV from that array. Bad idea?????


jcubbage:
Duplicates exist because we are combining tags from keys, equipment, and workstations for a given space.

jcubbage:
image

jcubbage:
image

jcubbage:
> looks good, so long as tags is never null. I'm going to assume it's not and 👍

Concat_WS yeilds empty string if all the values being concatenated are null.

PR #505, opened 2019-02-19
Lh/permissions

laholstege:
image

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
@@ -45,6 +45,9 @@ export default class AccessContainer extends React.Component<IProps, IState> {
         };
     }
     public async componentDidMount() {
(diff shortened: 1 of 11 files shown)


Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
```diff
@@ -45,6 +45,9 @@ export default class AccessContainer extends React.Component<IProps, IState> {
         };
     }
     public async componentDidMount() {
+        if (!PermissionsUtil.canViewAccess(this.context.permissions)) {
+            return <Denied viewName="Access" />;
```
srkirkland: @laholstege wait this can't actually work right?  You can't return a component from the did mount method, it's not `render()`.  What does this do (other than just return)?

laholstege: oops

Issue #497, opened 2019-02-14
On the Serial table, link to key. On Models, add table of spaces.


jcubbage:
That way you can see what rooms a person has access to.

PR #511, opened 2019-02-20
add link to key details

laholstege:
#497 i think?

image


laholstege:
@jSylvestre it would be nice to have an easy way to see all the spaces a person has access to, but i'm not sure if it's worth it either. 
we'll see what @srkirkland thinks

srkirkland:
@laholstege @jSylvestre Interesting idea, it'd have to be kinda within the keys area on the person details page.  I think this option is good either way -- even if we add a list of spaces we'll still want a quick link to the key itself.

PR #517, opened 2019-02-20
Jsc/tags report

### LINKED ISSUE #515: Tags report
Report on all the current tags in the system (so non-DA can see them).

AADI 2/20/19

jcubbage:
Closes #515 
And correct auth on wrong action.


jcubbage:
But not any of the OTHER roles. Jeremy requested we create this so non-admins can discover the tags the team has. He objected to the typeahead tags box. Thought it should either be a drop down multi select or create this report.

jSylvestre:
image

jcubbage:
@jSylvestre is right. Tags are a drop down pick-list already.

PR #520, opened 2019-02-21
Jcs/tag comma check

jSylvestre:


Keas.Mvc/Controllers/TagsManagerController.cs
```diff
@@ -38,11 +38,28 @@ public ActionResult Create()
         // POST: Tags/Create
         [HttpPost]
         public async Task<ActionResult> Create( Tag newTag)
-        {
+        {
+            var team = await _context.Teams.FirstAsync(t => t.Slug == Team);
```
srkirkland: Unrelated to your PR, but we should probably just throw this in a cache somewhere.  We look up the team all the time and it'd be really easy to cache into a service.

Keas.Mvc/Controllers/TagsManagerController.cs
```diff
@@ -38,11 +38,28 @@ public ActionResult Create()
...
+            if (ModelState.IsValid)
+            {
+                if (newTag.Name.Contains(","))
```
srkirkland: Instead why don't we validate w/ a regex in case we want to add other stuff.  Then we can just do normal regex validation on the name, without having to check.  Of course our regex could just check for commas, but I think we might actually want to just only allow `\w` alpha numeric

Issue #527, opened 2019-02-27
Equipment Type


jSylvestre:
Use equipment type.
Dropdown list
Have it show/hide some fields

PR #528, opened 2019-02-27
Equipment Type

jSylvestre:
Issue #527
Just make it editable first

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
@@ -65,7 +65,7 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
 
                     </div>
                     <div className="form-group">
-                        <label>Serial Number</label>
+                        <label>Serial Number / Identifier</label>
                         <input
                             type="text"
                             className="form-control"
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
@@ -179,13 +179,16 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
     };
 
     private _shouldShowForType(type: string, prop: string) {
-        if (prop !== "Card") {
-            return true;
-        }
-        if (type === "Make") {
+        if (prop === "Make") {
+            if (type !== "Card") {
+                return true
+            }
             return false;
         }
-        if (type === "Model") {
+        if (prop === "Model") {
+            if (type !== "Card") {
+                return true
+            }
             return false;
         }
 
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
@@ -18,6 +18,7 @@ interface IProps {
 
 export default class EquipmentEditValues extends React.Component<IProps, {}> {
     public render() {
+        const typeValue = this.props.selectedEquipment.type ? this.props.selectedEquipment.type : "Default";
         return (
             <div>
                 {this.props.disableEditing && this.props.openEditModal && (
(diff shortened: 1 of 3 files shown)


jSylvestre:
image

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -17,7 +17,8 @@ interface IProps {
 }
 
 export default class EquipmentEditValues extends React.Component<IProps, {}> {
-    public render() {
+    public render() {
+        const typeValue = this.props.selectedEquipment.type ? this.props.selectedEquipment.type : "Default";
```
srkirkland: @jSylvestre usually we use this syntax `const typeValue = this.props.selectedEquipment.type || "Default";`

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -45,9 +46,26 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
...
                         <FormFeedback>Item name is required</FormFeedback>
-                    </FormGroup>
+                    </FormGroup>
                     <div className="form-group">
-                        <label>Serial Number</label>
+                        <label>Type</label>
+                        <select
+                            className="form-control"
+                            value={typeValue}
+                            onChange={e =>this.props.changeProperty("type", e.target.value)}
```
srkirkland: Space after the =>.  We need to get you a tslinter in Visual Studio!

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -79,7 +98,9 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
                             }
                             onChange={e => this.props.changeProperty("make", e.target.value)}
                         />
-                    </div>
+                    </div>
+                    )}
+                    {this.props.selectedEquipment.type !== "Card" && (
```
srkirkland: "in the future" this would be better not inline w/ markup.  Maybe a method like `shouldShowForType(type, prop)` would be good, and that method would return true/false w/o having to change multiple places inline.

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -112,6 +112,12 @@ public async Task<IActionResult> Create([FromBody]Equipment equipment)
                    var space = await _context.Spaces.SingleAsync(x => x.RoomKey == equipment.Space.RoomKey);
                     equipment.Space = space;
                 }
+
+                if (equipment.Type.Equals("Card", StringComparison.OrdinalIgnoreCase))
```
srkirkland: Do we need to do this?  What happens in the other cases, some null error?

jSylvestre: @srkirkland We don't need it, but I figured we should clear out any values in case they are displayed somewhere else.

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
jSylvestre: @srkirkland Changed. But just monkey see monkey do:
https://github.com/ucdavis/Keas/blob/master/Keas.Mvc/ClientApp/components/Keys/KeySerialEditValues.tsx#L25

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -176,5 +176,19 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
 
     private _selectSpace = (space: ISpace) => {
         this.props.changeProperty("space", space);
+    };
+
+    private _shouldShowForType(type: string, prop: string) {
+        if (prop !== "Card") {
+            return true;
+        }
+        if (type === "Make") {
```
srkirkland: Throw this into the else, since this only applies to prop === "Card"

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -176,5 +176,19 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
...
+    };
+
+    private _shouldShowForType(type: string, prop: string) {
+        if (prop !== "Card") {
+            return true;
+        }
+        if (type === "Make") {
+            return false;
+        }
+        if (type === "Model") {
```
srkirkland: shouldn't type == "Card" and prop == "Make"?

Keas.Mvc/Controllers/Api/EquipmentController.cs
srkirkland: hmm.. I don't like the double logic of it here, which I think will get out of hand pretty quickly when more types are added.  Either we should leave it alone (my thought) or in react, when type is changed we clear out fields which shouldn't apply (similar to your shouldShow method)

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -154,5 +176,22 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
 
     private _selectSpace = (space: ISpace) => {
         this.props.changeProperty("space", space);
+    };
+
+    private _shouldShowForType(type: string, prop: string) {
+        if (prop === "Make") {
+            if (type !== "Card") {
+                return true
```
srkirkland: we want semi colons!

jSylvestre: ;)

PR #530, opened 2019-02-27
Jcs/relax iam lookup

jSylvestre:


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.EntityFrameworkCore;
 using System.Text;
 using ietws.PPSDepartment;
+using Microsoft.EntityFrameworkCore.Design;
```
srkirkland: @jSylvestre what is this doing?

jSylvestre: @srkirkland Don't know. It isn't used so I created another PR to clean that and a couple other things up pr https://github.com/ucdavis/Keas/pull/531

PR #548, opened 2019-03-12
How about this

jcubbage:
Basically same as @jSylvestre  suggested fix.

### DIFF
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -227,7 +227,14 @@ namespace Keas.Mvc.Services
                         newpeople += personResult.peopleCount;
                         if (personResult.Person == null)
                         {
-                            warning.Append($"IAM ID {id.IamId} failed to save.");
+                            if (kerbResults.ResponseData.Results.Length > 0)
+                            {
+                                warning.Append($" IAM ID: {id.IamId}, Kerb Id: {kerbResults.ResponseData.Results.First().UserId} failed to save.");
+                            }
+                            else
+                            {
+                                warning.Append($" IAM ID {id.IamId} failed to save.");
+                            }
                         }
                     }
                 }
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -28,7 +28,7 @@ namespace Keas.Mvc.Services
     {
         private readonly AuthSettings _authSettings;
         private readonly ApplicationDbContext _context;
-    
+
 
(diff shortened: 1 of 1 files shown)


jcubbage:
image

Keas.Mvc/Services/IdentityService.cs
```diff
@@ -199,6 +225,10 @@ public async Task<string> BulkLoadPeople(string ppsCode, string teamslug)
                     {
                         var personResult = await GetOrCreatePersonFromKerberos(kerbResults.ResponseData.Results[0].UserId, team.Id);
                         newpeople += personResult.peopleCount;
+                        if (personResult.Person == null)
+                        {
+                            warning.Append($"IAM ID {id.IamId} failed to save.");
```
jSylvestre: @jcubbage IAM Id isn't supper useful to me, so I'd add kerb as well, so I'd change it to:
warning.Append($"IAM ID: {id.IamId}, Kerb Id: {kerbResults.ResponseData.Results.First().UserId}  failed to save.");

Keas.Mvc/Services/IdentityService.cs
```diff
@@ -199,6 +225,10 @@ public async Task<string> BulkLoadPeople(string ppsCode, string teamslug)
...
+                        }
                     }
```
jSylvestre: @jcubbage Add an else with a warning here, but you can probably only have the IAM id as the kerb may not have been found. This should be the else to the `if (kerbResults.ResponseData.Results.Length > 0 && contactResult.ResponseData.Results.Length > 0)`

jSylvestre: @jcubbage That will not catch IAM users that PPS returns, but that we don't find.
You would need it on the else of this if (If we want to report these)
`                    var kerbResults = await clientws.Kerberos.Search(KerberosSearchField.iamId, id.IamId);
                    var contactResult = await clientws.Contacts.Get(id.IamId);

                    if (kerbResults.ResponseData.Results.Length > 0 && contactResult.ResponseData.Results.Length > 0)
                    {
                        var personResult = await GetOrCreatePersonFromKerberos(kerbResults.ResponseData.Results[0].UserId, team.Id);
                        newpeople += personResult.peopleCount;
                        if (personResult.Person == null)
                        {
                            if (kerbResults.ResponseData.Results.Length > 0)
                            {
                                warning.Append($" IAM ID: {id.IamId}, Kerb Id: {kerbResults.ResponseData.Results.First().UserId} failed to save.");
                            }
                            else
                            {
                                warning.Append($" IAM ID {id.IamId} failed to save.");
                            }
                        }
                    }`

PR #554, opened 2019-03-14
This fixes some of the people not found.

jSylvestre:
The IAM lookup still needs to return a loginId

### DIFF
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -187,7 +187,7 @@ namespace Keas.Mvc.Services
                 FirstName = ucdKerbPerson.DFirstName,
                 LastName = ucdKerbPerson.DLastName,
                 Id = ucdKerbPerson.UserId,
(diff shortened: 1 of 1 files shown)


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -187,7 +187,7 @@ public async Task<User> GetByKerberos(string kerb)
                 FirstName = ucdKerbPerson.DFirstName,
                 LastName = ucdKerbPerson.DLastName,
                 Id = ucdKerbPerson.UserId,
-                Email = ucdContact.Email,
+                Email = ucdContact.Email ?? (ucdKerbPerson.UserId != null ? $"{ucdKerbPerson.UserId}@ucdavis.edu" : null),
```
srkirkland: Just use an IF statement here, nesting a null coalescence `??` and a tertiary together is not easy to read.  `var email = ucdContext.Email;  if (email == null) { ... }` that kind of thing so we can tell what's happening

PR #556, opened 2019-03-14
Try to get IAM name on fail (BulkLoadPeople)

jSylvestre:
image

### DIFF
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -271,9 +271,9 @@ namespace Keas.Mvc.Services
                                 extraName = $"({iamInfoResults.ResponseData.Results.First().OFullName}) ";
                             }
                         }
-                        catch (Exception e)
+                        catch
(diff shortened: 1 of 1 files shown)


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -260,8 +260,23 @@ public async Task<string> BulkLoadPeople(string ppsCode, string teamslug)
...
+                        try
+                        {
+                            var iamInfoResults = await clientws.People.Get(id.IamId);//  //.Kerberos.Search(KerberosSearchField.iamId, id.IamId);
+                            if (iamInfoResults.ResponseData.Results.Length > 0)
+                            {
+                                extraName = $"({iamInfoResults.ResponseData.Results.First().OFullName}) ";
+                            }
+                        }
+                        catch (Exception e)
+                        {
```
srkirkland: If you aren't catching you shouldn't declare the exception var.  So like `catch {}` only

PR #561, opened 2019-03-20
Jcs/person notification

jSylvestre:


Keas.Core/Services/IEmailService.cs
```diff
@@ -100,6 +102,82 @@ public async Task SendTeamExpiringMessage(int teamId, ExpiringItemsEmailModel mo
...
 
+        public async Task SendPersonNotification()
+        {
+            if (_emailSettings.DisableSend.Equals("Yes", StringComparison.OrdinalIgnoreCase))
+            {
+                Log.Information("Email Sending Disabled");
+                return;
+            }
+
+            var personEmails = await _dbContext.PersonNotifications.Where(a => a.Pending && a.SendEmail).Select(a => a.NotificationEmail).Distinct().ToListAsync();
```
srkirkland: not too important but you can do `asNoTracking` and you can also to .ToArrayAsync(), since it doesn't actually need to be a list (lists are for when you are going to modify what is in that list, like add/remove stuff).

Issue #560, opened 2019-03-20
No Team Validation for Home Index


jSylvestre:
https://peaks.ucdavis.edu/this-is-not-valid/home/index

PR #563, opened 2019-03-20
Use an ErrorMessage and redirect to page to select team

jSylvestre:


jSylvestre:
Close #560

jcubbage:
Do we need to do the same sort of thing on ALL the other pages? Should this be part of the SuperController? Does that become too costly unless we do caching?

jSylvestre:
@jcubbage I'd say just here.

PR #565, opened 2019-03-21
Jcs/prevent removing self

### LINKED ISSUE #551: Deleting Access And People
So... You probably shouldn't be able to remove your own DA access for a team (Must get another DA to do it).

Also, we should restrict deleting a person if there any any admin roles (Otherwise thinks break in the notifications)

jSylvestre:
Close #551


jcubbage:
Sounds like natural consequences in action to me.

jSylvestre:
@srkirkland Yes. Because if someone removes their own access, I have to go in an fix it.
In addition, this can happen if you delete the "person".

PR #569, opened 2019-03-28
Jcs/more tracks

jSylvestre:
image

### DIFF
--- a/Keas.Mvc/Controllers/Api/KeySerialsController.cs
+++ b/Keas.Mvc/Controllers/Api/KeySerialsController.cs
@@ -149,7 +149,7 @@ namespace Keas.Mvc.Controllers.Api
             key.Serials.Add(keySerial);
             await _context.SaveChangesAsync();
             
-            //await _eventService.TrackCreateKeySerial(keySerial);
+            await _eventService.TrackCreateKeySerial(keySerial);
 
             return Json(keySerial);
         }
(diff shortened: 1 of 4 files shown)


jSylvestre:
I'm using the key title, want me to use the key code instead?

jSylvestre:
> in the future our services shouldn't be doing _dbContext.SaveChanges(). The calling code should be responsive for triggering save (or not). The way it currently is isn't the best for performance or ability to rollback in case of errors.

That is something for future @jcubbage :)

jcubbage:
> I'm using the key title, want me to use the key code instead?

I think we should always be using asset.Title. For keys that returns name if available, code if not. KeySerials get name then number. For KeySerial we may want to extend further with FullName => {key.Title}-{keyserial.Title}?

jcubbage:
> in the future our services shouldn't be doing _dbContext.SaveChanges(). The calling code should be responsive for triggering save (or not). The way it currently is isn't the best for performance or ability to rollback in case of errors.

Does that mean our controllers should be saving changes twice? We need the full object in the event service, so do we start doing this:

await _eventService.TrackAssignAccess(accessAssignment, Team);

srkirkland:
@jcubbage I don't know enough about the specific problems but I would think anything you need in the event service you already have without saving.

jcubbage:
> @jcubbage I don't know enough about the specific problems but I would think anything you need in the event service you already have without saving.

@srkirkland You were right. I thought the object had to be saved before calling the eventService...but I was mistaken.
Did new PR with all the saveChanges pulled from history & notification services. Can either merge with this, or review/approve separate.

PR #570, opened 2019-03-28
Lh/key serial table filtering

### LINKED ISSUE #566: Sorting and Searching on Key Serials table
The Key serials table can be large. There is no searching or sorting for this table.

image

==================
For ticket INC0757405

There is a work around for this but it is very time consuming from an admin perspective.

We have hundreds of keys of the same key code. We need the ability to find one particular key (key code plus sn) quickly. Can you add a sort feature to the page

key/details...

E.g., Physics key code AB. When I find that key and click on Details, a long list in tabular format with columns Code, SN, Status, Assigned To, Expiration, Actions. I'd like to be able to sort on any of those headers, but in particular SN.

Also, the filter for Unassigned/Assigned/Any needs to be on the Key Details list page. This filter where it sits now seems to be out of place.

laholstege:
adds react table on keys/details serial list with sort/filter
image
closes #566

### DIFF
--- a/Keas.Mvc/ClientApp/components/Keys/KeySerialTable.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/KeySerialTable.tsx
@@ -89,15 +89,20 @@ export default class KeySerialTable extends React.Component<IProps, {}> {
                     },
                     {
                         Header: "Assignment",
-                        accessor: (keySerial: IKeySerial) => keySerial.keySerialAssignment ? keySerial.keySerialAssignment.person.name : null,
+                        accessor: (keySerial: IKeySerial) =>
+                            keySerial.keySerialAssignment
+                                ? keySerial.keySerialAssignment.person.name
+                                : null,
                         className: "word-wrap",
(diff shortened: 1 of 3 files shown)


Keas.Mvc/ClientApp/components/Keys/KeySerialTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+                                style={{ width: "100%" }}
+                                value={filter ? filter.value : "all"}
+                            >
+                                <option value="all">Show All</option>
+                                <option value="unassigned">Unassigned</option>
+                                <option value="expired">Expired</option>
+                                <option value="unexpired">All Unexpired</option>
+                                <option value="3weeks">Expiring within 3 weeks</option>
+                                <option value="6weeks">Expiring within 6 weeks</option>
+                            </select>
```
srkirkland: We use this same pattern quite a bit, including repeating these exact options.  Next time we change a component let's extract to a new shared expiration filter component.

PR #573, opened 2019-03-28
let key table search work for full key code+SN

laholstege:
image


jSylvestre:
It seems kind of strange that this matches anything:

image

laholstege:
@srkirkland thought's on @jSylvestre 's comment? i'm not there's a way to match anything containing the key code and also filter out garbage

PR #580, opened 2019-04-02
Jsc/equipment type search

### LINKED ISSUE #553: Add equipment type as searchable property
Either as filter (like tags & attributes) or add as new sortable/searchable column on equipment table.

jcubbage:
Closes #553 
That was too easy?!?!?!?!

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -192,7 +192,7 @@ export default class EquipmentContainer extends React.Component<IProps, IState> 
             }
             if (this.state.equipmentTypeFilters.length > 0) {
                 filteredEquipment = filteredEquipment.filter(x => 
-                    this._checkEquipmentTypeFilters(x this.state.equipmentTypeFilters)
+                    this._checkEquipmentTypeFilters(x, this.state.equipmentTypeFilters)
                 );
             };
             return (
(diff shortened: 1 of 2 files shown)


Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -187,6 +190,11 @@ export default class EquipmentContainer extends React.Component<IProps, IState>
                     this._checkAttributeFilters(x, this.state.attributeFilters)
                 );
             }
+            if (this.state.equipmentTypeFilters.length > 0) {
+                filteredEquipment = filteredEquipment.filter(x => 
+                    this._checkEquipmentTypeFilters(x, this.state.equipmentTypeFilters)
```
srkirkland: something to think about -- you don't need to pass `this.state.equipmentTypeFilters`, since that method is part of this class and already has access to `this.state`.  If you did, this line could become `filteredEquipment.filter(this._checkEquipmentTypeFilters)`.

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -426,6 +440,14 @@ export default class EquipmentContainer extends React.Component<IProps, IState>
         }
     };
 
+    private _filtersEquipmentType = (filters: string[]) => {
```
srkirkland: I think you have an extra `s` in here, should be `filterEquipmentType`?

Issue #546, opened 2019-03-12
Reports & API enhancements


srkirkland:
Need reports for at least each of PEAKS.  List my people, list equipment, list of workstations, etc.

API should also be extended to return the same information (or superset) as each report.  If possible, it would be great to even drive the report tables off of the API calls directly, since that'll give us parity between reports and the API.

jcubbage:
- [ ] People - same as list() except replace counts with actual asset objects.

- [ ] Equipment - list()

- [ ] Access - list()

Keys:

PR #581, opened 2019-04-02
Jcs/reports20190402

jSylvestre:
This PR includes the api for the same info

### DIFF
--- a/Keas.Mvc/Controllers/FeedController.cs
+++ b/Keas.Mvc/Controllers/FeedController.cs
@@ -20,7 +20,7 @@ namespace Keas.Mvc.Controllers
 
          public FeedController(ApplicationDbContext context, IReportService reportService)
          {
-             this._context = context;
+             _context = context;
              _reportService = reportService;
          }
         public async Task<IActionResult> TeamFeed(Guid id, string includeSpace)
--- a/Keas.Mvc/Controllers/FeedController.cs
+++ b/Keas.Mvc/Controllers/FeedController.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Keas.Core.Domain;
+using Keas.Mvc.Services;
 
 namespace Keas.Mvc.Controllers
 {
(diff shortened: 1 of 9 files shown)


jSylvestre:
image

image

jSylvestre:
Part of issue #546

jcubbage:
The only thing I'd say is our list of API links at TEAM/TeamAdmin/RegenerateApiCode is going to get really big. We should pull that out to a new page to list out all the feeds teams will eventually have.

Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -8,18 +8,21 @@
...
+         private readonly IReportService _reportService;
 
-        public FeedController(ApplicationDbContext context)
-        {
-            this._context = context;
-        }
+         public FeedController(ApplicationDbContext context, IReportService reportService)
+         {
+             this._context = context;
+             _reportService = reportService;
```
srkirkland: why use `this._context` one line up but not `this._reportService` on this line?

Keas.Mvc/Controllers/ReportController.cs
```diff
@@ -19,11 +19,13 @@ public class ReportController : SuperController
...
 
         private readonly ISecurityService _securityService;
+        private readonly IReportService _reportService;
 
-        public ReportController(ApplicationDbContext context, ISecurityService securityService)
+        public ReportController(ApplicationDbContext context, ISecurityService securityService, IReportService reportService)
         {
             this._context = context;
             this._securityService = securityService;
+            _reportService = reportService;
```
srkirkland: same here -- you should change to make them match

Keas.Mvc/Services/ReportService.cs
```diff
@@ -0,0 +1,63 @@
...
+
+
+        public async Task<IList<WorkstationReportModel>> WorkStations(Team team, string teamSlug)
+        {
+            if (team == null)
+            {
+                team = await _context.Teams.SingleAsync(a => a.Slug == teamSlug);
+            }
+
+            return await _context.Workstations.Include(a => a.Space).Include(a => a.Assignment).ThenInclude(a => a.Person).IgnoreQueryFilters().AsNoTracking().Where(a => a.TeamId == team.Id).Select(a => new WorkstationReportModel
```
srkirkland: try getting rid of the `.Include(...)` parts of this and run again.  I don't think you'll need them and I think they might be grabbing more info than you need.  You don't need the `.Include()` if you are projecting out with a `.Select()`

Keas.Mvc/Views/Report/Index.cshtml
```diff
@@ -63,5 +63,18 @@
     <div class="card-content">
         <a asp-action="PersonActions" class="btn btn-link">Actions taken against people in team <i class="fas fa-chevron-right fa-sm" aria-hidden="true"></i></a>
     </div>
-</div>
+</div>
+ @if ((await AuthorizationService.AuthorizeAsync(User, AccessCodes.Codes.SpaceMasterAccess)).Succeeded)
```
srkirkland: this is ok but we should come up with a better way of authorizing reports, or cache this.  Currently the report page is going to be doing 2 queries per report so we are probably up to 10 already, just to figure out what the user can access.  One idea is to query for the user's roles server side and then use that to check if user isInRole inside the page.  Something to keep in mind anyway

Keas.Mvc/Controllers/FeedController.cs
jSylvestre: Don't need the this. I'll remove that in both places 
Blame @jcubbage  :)

Issue #588, opened 2019-04-05
From Help Ticket


jSylvestre:
INC0763158
Two feature requests:
 
1) Allow people API to pull all People Data: supervisor, category, etc
2) Create another access role for editing people. Because admin assistants often need to update people's profiles (especially after importing 300 people with no contact information), we are reluctant to give them Departmental Admin access to do so.

jcubbage:
2) Ooooohhh.....PeopleMaster seems like a really bad name!

jSylvestre:
@jcubbage 1 shouldn't be a problem, Just add those fields to the model and populate them?

jcubbage:
@jSylvestre  has 1) covered.
@jcubbage  will take 2)

PR #591, opened 2019-04-05
Add fields to People feeds

jSylvestre:
Also show inactive people (with a new field for Active)
Added NoTracking to query


jSylvestre:
part of issue #588

PR #593, opened 2019-04-05
Jsc/person manager

jcubbage:
Part 2 of #588 covered in this PR.
Added new role type "PersonManager". Updated security policies, controller permissions, React permissions.
Also removed SystemAdmins (Admins) from permissions in React page. We had removed their access in controllers, but not react.


jcubbage:
Pending @jSylvestre updating tests (check number of codes, etc).

PR #596, opened 2019-04-08
Jcs/key feed1

jSylvestre:
Part of #546

### DIFF
--- a/Keas.Mvc/Controllers/FeedController.cs
+++ b/Keas.Mvc/Controllers/FeedController.cs
@@ -75,5 +75,17 @@ namespace Keas.Mvc.Controllers
             return Json(await _reportService.AccessList(null, Team));
         }
 
+        public async Task<IActionResult> KeyFeed(Guid id)
+        {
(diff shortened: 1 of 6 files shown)


Keas.Mvc/Controllers/FeedController.cs
```diff
@@ -73,6 +73,18 @@ public async Task<IActionResult> AccessFeed(Guid id)
             }
 
             return Json(await _reportService.AccessList(null, Team));
+        }
+
+        public async Task<IActionResult> KeyFeed(Guid id)
+        {
+            var validKey = await _context.Teams.Where(t => t.Slug == Team && t.ApiCode != null && t.ApiCode == id).AnyAsync();
```
srkirkland: What does the `t.ApiCode != null` do for you here?  I can't think of a case where this clause makes any difference.

jSylvestre: Just copy and paste @jcubbage code. Not nullable, so I think you are correct.

jcubbage: @jSylvestre Team.ApiCode is nullable (Guid?). This prevent someone from pulling the feed on a team that has not set their API code.

Issue #577, opened 2019-04-02
Bulk Edit People


jSylvestre:
From a help ticket, not sure if we want to do this:

4. Is there a way to bulk edit people? For example, I might want to edit a whole bunch of people to have the same category (staff), tag (CRU), or end date. Is there any way to facilitate this without having to click into each person individually, click edit, change the fields individually, then click Go?

jcubbage:
I'm not sure how they would identify who they want to edit. Maybe if you wanted to tag all the staff with the same tag. But after initial load, how are you going to decide who needs a particular update?

jSylvestre:
Scott suggested a non-react page that would list people with a checkbox and then have a few fields they could update.

PR #601, opened 2019-04-11
BulkEdit form

jSylvestre:
Issue #577

### DIFF
--- a/Keas.Core/Models/PersonCategories.cs
+++ b/Keas.Core/Models/PersonCategories.cs
@@ -25,16 +25,5 @@ namespace Keas.Core.Models
             Volunteer,
         };
 
-        public static List<string> BulkEditChoices = new List<string>
-        {
-            "-- Not Set --",
-            Faculty,
-            Staff,
-            AdminStaff,
-            GradStudent,
-            Undergrad,
-            Visitor,
-            Volunteer,
-        };
     }
 }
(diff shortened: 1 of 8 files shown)


jSylvestre:
image

jSylvestre:
image

image

Keas.Mvc/Controllers/TeamAdminController.cs
```diff
@@ -857,6 +859,154 @@ public IActionResult UploadEquipment()
...
+                            return View(model);
+                        }
+                    }
+                }
+                foreach (var person in persons)
+                {
+                    
+                    if (model.UpdateCategory)
+                    {
+                        if (model.Category == "-- Not Set --")
```
srkirkland: a little too magic stringy here

Keas.Core/Models/PersonCategories.cs
```diff
@@ -0,0 +1,40 @@
...
+            AdminStaff,
+            GradStudent,
+            Undergrad,
+            Visitor,
+            Volunteer,
+        };
+
+        public static List<string> BulkEditChoices = new List<string>
+        {
+            "-- Not Set --",
```
srkirkland: This is always one of those things where I'd rather leave this out (since it's not a model layer concern) and then have the controller create the SelectList with the bulkEditChoices plus a default choice and pass that to the view.  Would also make it easier to later compare without having to retype `"-- Not Set --"` and then any chance to those default value will need to be changed in 2 places.

PR #611, opened 2019-04-16
Add a supervisor column

### LINKED ISSUE #362: Supervisor on front page, perhaps as mouseover
More info as a mouseover on the people tab.
From the CAO/IT Lead meeting - 1/10/19

jSylvestre:
close #362


jSylvestre:
Not sure is we want it like this.

image

jSylvestre:
> why not have a popover like the issue says? I guess this is fine, though I like to limit the number of columns. What percentage of people in the system have supervisor populated? I'd say if it's less than half we probably shouldn't have a column

@srkirkland Because this was the limit of my React ability :) As to the statistics for use, those would not be accurate as people can get added easy with bulk load, but may not have been edited yet.
If someone wants to d a popover, I'm fine with that, but I kind of like the filtering/sorting this provides.

PR #617, opened 2019-04-18
Jcs/session

jSylvestre:
Cache calls to team permissions for the security service


Keas.Mvc/Services/SessionManagementService.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading.Tasks;
+using Keas.Core.Data;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Rewrite.Internal.ApacheModRewrite;
+using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
+
+namespace Keas.Mvc.Services
+{
+    public interface ISessionManagementService
```
jpknoll: pick a more specific name

RolesSessionsManager

Keas.Mvc/Services/SessionManagementService.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Will return empty array if no roles.
+        /// </summary>
+        /// <param name="slug"></param>
+        /// <returns></returns>
+        public async Task<string[]> GetTeamRoleNames(string slug)
+        {
+            var userId = _contextAccessor.HttpContext.User.Identity.Name;
+            if (_contextAccessor.HttpContext.Session.GetString($"UserId:{userId}") != userId)
+            {
+                _contextAccessor.HttpContext.Session.Clear(); //Abandon?
```
jpknoll: clears the entire session, logout?

jSylvestre:
image

Keas.Mvc/Services/RolesSessionsManager.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Threading.Tasks;
+using Keas.Core.Data;
+using Microsoft.AspNetCore.Http;
+using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
+using Serilog;
+
+namespace Keas.Mvc.Services
+{
+    public interface IRolesSessionsManager
```
srkirkland: Something drastic w/ naming here -- how about just `IRolesManager`?  The interface shouldn't care if the implementation is actually done via session or cache or cookies or w/e.  Then we'd want to change the `ClearSessionRoles` method to just `ClearRoles` so it's more generic.

Keas.Mvc/Services/RolesSessionsManager.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+        Task<string[]> GetTeamRoleNames(string slug);
+        Task<string[]> GetTeamOrAdminRoleNames(string slug);
+        Task<string[]> GetSystemRoleNames();
+        void ClearSessionRoles();
+
+    }
+
+    public class RolesSessionsManager : IRolesSessionsManager
```
srkirkland: For the implementation, i don't like the double plural.  There aren't multiple sessions so maybe just `RolesSessionManager`?

Keas.Mvc/Services/RolesSessionsManager.cs
```diff
@@ -0,0 +1,170 @@
...
+            var userId = _contextAccessor.HttpContext.User.Identity.Name;
+            var systemRole = roleContainer.SystemRoles;
+            if (systemRole == null)
+            {
+                systemRole = new Team
+                {
+                    TeamName = "System",
+                    TeamRoles = await _dbContext.SystemPermissions
+                        .AsNoTracking()
+                        .Where(a => a.UserId == userId).Select(a => a.Role.Name).Distinct().ToArrayAsync(),
```
srkirkland: @jSylvestre It's safe to leave but I'm not sure you need asNoTracking() if you are going to .Select() project out.  There is nothing to track anyway

Keas.Mvc/Services/RolesSessionsManager.cs
jpknoll: Yeah, this is the right direction.

Keas.Mvc/Services/RolesSessionsManager.cs
jpknoll: My concern around RolesManager is that name already exists for the EF Roles Manager.

Keas.Mvc/Services/RolesSessionsManager.cs
jpknoll: Yeah, if the query is returning just a list of strings, he can probably leave it out. 

I'd even consider raw SQL for this part of the application.

jSylvestre:
close issue #618

Keas.Mvc/Services/TeamsManager.cs
```diff
@@ -0,0 +1,192 @@
...
+
+namespace Keas.Mvc.Services
+{
+    public interface ITeamsManager
+    {
+        Task<string[]> GetTeamRoleNames(string slug);
+        Task<string[]> GetTeamOrAdminRoleNames(string slug);
+        Task<string[]> GetSystemRoleNames();
+        Task<Dictionary<string, string>> GetMyTeams();
+        void ClearTeamSession();
```
srkirkland: still don't like this having the session as part of the name, but it's not that bad

Keas.Mvc/Services/TeamsManager.cs
```diff
@@ -0,0 +1,192 @@
...
+            public List<Team> Teams { get; set; }
+
+            public Team SystemRoles { get; set; }
+            public Dictionary<string, string> TeamDictionary { get; set; }
+        }
+
+        public class Team
+        {
+            public string TeamName { get; set; }
+            public string[] TeamRoles { get; set; }
```
srkirkland: Generally I don't think it's necessary to add the classname to the properties of that class, like I'd rather you have `roles.team.name` than `roles.team.teamName`.

PR #622, opened 2019-04-26
Check System Admin roles for menu.

### LINKED ISSUE #621: Super Admin Menu
Toasted the Super Admin menu for Team Roles with the cache changes

jSylvestre:
This is cached, but it also means every user now checks this.
It makes it easier for me as an admin by giving me visual queues and adds a link so I don't have to remember the TeamAdmin/RoledMembers link when setting up new teams

Close #621


jSylvestre:
Meh, better way is #623

PR #626, opened 2019-05-02
Limit viewed links with permissions

### LINKED ISSUE #514: Hide tabs if no access
Update react assetNav to only show tabs that you have permissions to see.

AADI 2/20/19

jcubbage:
Closes #514 
Hey, look! I did another thing in React!
TSLint said page had imports that were unused....I think those were ok to get rid of.


srkirkland:
So this means we won't get the PEAKS tabs for everyone?

PR #631, opened 2019-05-31
Jcs/group teams1

jSylvestre:


Keas.Core/Domain/GroupXTeam.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Keas.Core.Domain
+{
+    public class GroupXTeam
+    {
+        [Key]
+        public int Id { get; set; }
+
+        [Required]
+        public Group Key { get; set; }
```
jSylvestre: Oops

Keas.Mvc/Controllers/ConfirmController.cs
```diff
@@ -177,19 +178,27 @@ public async Task<IActionResult> AcceptAll()
             var people = await _context.People.Where(p => p.User == user).Select(a => a.Team).AsNoTracking().ToArrayAsync();
             var teamAdmins = await _context.TeamPermissions.Where(tp => tp.User == user).Select(a => a.Team).AsNoTracking().ToArrayAsync();
             var teams = people.Union(teamAdmins, new TeamComparer());
-            if(teams.Count() == 0){
+
+            var groupPermissions = await _context.GroupPermissions.Include(a => a.Group).ThenInclude(a => a.Teams).ThenInclude(a => a.Team).AsNoTracking().Where(a => a.UserId == user.Id).ToListAsync();
```
srkirkland: Why do we need to include teams->team?  Shouldn't we just care which groups they have access to and ignore teams that are part of those groups?

Keas.Mvc/Controllers/TeamController.cs
```diff
@@ -105,5 +110,141 @@ private bool TeamExists(int id)
...
+
+            var groupPermission = new GroupPermission{ GroupId = group.Id, UserId = user.Id };
+
+            _context.GroupPermissions.Add(groupPermission);
+            await _context.SaveChangesAsync();
+            Message = "User " + user.Name + " has been added.";
+            return RedirectToAction("GroupDetails", "Team", new { id = model.Group.Id });
+        }
+
+        public async Task<IActionResult> RemoveGroupie(int id, int groupId)
```
srkirkland: why remove grupie but not add groupie? or does the group user get added but then they become groupies and have to be removed?

Keas.Mvc/Keas.Mvc.csproj
```diff
@@ -31,6 +31,7 @@
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Data\" />
+    <Folder Include="Pages\" />
```
srkirkland: why?

Keas.Mvc/Controllers/ConfirmController.cs
jSylvestre: @srkirkland Good point. Was doing it a different way at first. I'll fix tomorrow.

Keas.Mvc/Controllers/TeamController.cs
jSylvestre: No one should add groupies :) Was just having fun. Can change it.

Keas.Mvc/Keas.Mvc.csproj
jSylvestre: Oops, didn't see that. It is where the view got created. No idea why.

jSylvestre:
Test and Prod databases have been updated

PR #637, opened 2019-06-05
Trim and ignore case

jSylvestre:


Keas.Mvc/Controllers/SystemController.cs
```diff
@@ -36,7 +36,7 @@ public async Task<IActionResult> Emulate(EmulateUserViewModel model)
         {
             //Log.Information("Emulation attempted for {username}", username);
 
-            var user = await _context.Users.SingleOrDefaultAsync(u => u.Email == model.UserEmail || u.Id == model.UserEmail);
+            var user = await _context.Users.SingleOrDefaultAsync(u => u.Email.Equals(model.UserEmail.Trim(), StringComparison.OrdinalIgnoreCase) || u.Id.Equals(model.UserEmail.Trim(), StringComparison.OrdinalIgnoreCase));
```
srkirkland: @jSylvestre Maybe do the trim ahead of time and then pass in the trimmed value, instead of trimming twice inside the query.  I'm concerned EF won't generate the query properly with the trim inside (performance wise), but plus it'll be cleaner to just trim once and drop it in a var, since it's a server-side thing

Keas.Mvc/Controllers/SystemController.cs
```diff
@@ -35,8 +35,9 @@ public IActionResult Emulate()
         public async Task<IActionResult> Emulate(EmulateUserViewModel model)
         {
             //Log.Information("Emulation attempted for {username}", username);
+            var lookupVal = model.UserEmail.Trim();
 
-            var user = await _context.Users.SingleOrDefaultAsync(u => u.Email == model.UserEmail || u.Id == model.UserEmail);
+            var user = await _context.Users.SingleOrDefaultAsync(u => u.Email.Equals(lookupVal, StringComparison.OrdinalIgnoreCase) || u.Id.Equals(lookupVal, StringComparison.OrdinalIgnoreCase));
```
srkirkland: How will the second part of the OR ever be true?  Why would the user email match the u.Id value?

jSylvestre: @srkirkland We can lookup be kerb. This is a System Admin only page as well.

srkirkland: @jSylvestre if we can lookup by kerb then why is the variable called UserEmail?

jSylvestre: @srkirkland Probably copy and paste. But this functionality is in more than one place. I blame @jcubbage but it might have been me.

srkirkland: ok fine 🚢

PR #640, opened 2019-06-11
Remember React Table page size

### LINKED ISSUE #552: React not remembering show 100 rows
@emhenn Didn't like it, so complain to him :)

jSylvestre:
Close #552


srkirkland:
@jSylvestre I think this could do with going into a util or service file.  Then you'd just import the tableUtil.ts or whatever and do `getPageSize()` which would encapsulate all the logic.  I don't like having to put the key value in several different places, the same code is too repeated

jSylvestre:
@srkirkland Better? Is the proj file change ok?

PR #651, opened 2019-07-17
the modals escape issue fixed.

NabilFurmoli:


jcubbage:
Closes #652. Closes #651

Keas.Mvc/ClientApp/components/Access/AssignAccess.tsx
```diff
@@ -74,7 +74,7 @@ export default class AssignAccess extends React.Component<IProps, IState> {
                 >
                     <div className="modal-header row justify-content-between">
                         <h2>{this.props.selectedAccess || this.props.person ? "Assign Access" : "Add Access"}</h2>
-                        <Button color="link" onClick={this._closeModal}>
+                        <Button color="link" onClick={this._closeModal_using_icon}>
```
laholstege: @NabilFurmoli try to keep the naming on your functions consistent. The original function is private, so it's `_closeModal`, so your new function should be something like `_closeModalUsingIcon`, or name your new function `_confirmCloseModal`. Your new function might also just call `_closeModal` itself after confirming, so that you don't have reused logic.

Keas.Mvc/ClientApp/components/Keys/CreateKey.tsx
```diff
@@ -1,176 +1,197 @@
-import * as PropTypes from "prop-types";
```
laholstege: what are you changing here? line endings? formatting? try to make sure it's clear what's changed when you push files--the way it is now, it's hard to tell if you've changed anything

Keas.Mvc/ClientApp/components/Access/AssignAccess.tsx
NabilFurmoli: Ok sounds good I will update the function names.

laholstege:
looks good! :ship:

Issue #650, opened 2019-07-16
IS3 Classifications


jSylvestre:
So, we may have need for new required fields.
Perhaps have a flag that says IS3 Requirement that will show fields and or make them required?

Expand access to have more fields such as a URL to a password manager for cert details?

Think about adding a list serve email for some kinds of expiration? Maybe still assign it to a person, but have a generic email that we can send to as well.

* [x] Add Fields to DB
* [x] Add Fields to domain object
* [x] Add Fields to Form(s)
* [ ] ~Update any dapper queries~
* [x] Update CSV Import
* [x] Script to default data based on Type
* [x] Search Fields on equipment list? (Maybe have a ... and dialog for more search fields)
* [ ] Link on form to official page for what these values are?
* [x] Reports / Json Dumps

srkirkland:
https://security.ucop.edu/policies/institutional-information-and-it-resource-classification.html

@jSylvestre @laholstege So the idea would be we add two new fields for Equipment: "Protection Level" and "Availability Level".  They would be dropdowns w/ the values specified in that doc.

For all equipment types, we can optionally select one of the values for each field.  For some types they will be required (anything computer-related, specifically anything that can store data).

In the form for create/edit we should put it before "notes" so it's a first-class citizen

jSylvestre:
@srkirkland Have you thought about how you want to populate existing data if it is required? Or just have it required on the page?

srkirkland:
@jSylvestre well it can't be required in the DB because not all types will have info for those fields.  So I was just thinking required on the page, and thus old equipment would just be lazily updated as people next edit them.

jSylvestre:
* [ ] Add Fields to DB
* [ ] Add Fields to domain object
* [ ] Add Fields to Form(s)
* [ ] Update CSV Import
* [ ] Script to default data based on Type
* [ ] Search Fields on equipment list? (Maybe have a ... and dialog for more search fields)
* [ ] Link on form to official page for what these values are

jSylvestre:
Added filter in PR #666

PR #665, opened 2019-07-26
Jcs/is3 fields

jSylvestre:
Issue #650


Keas.Mvc/Controllers/TeamAdminController.cs
```diff
@@ -1090,6 +1095,26 @@ private Equipment CreateEquipment(EquipmentImport r, Team team, EquipmentImportR
...
+
+                if (typesWithPartectionAndAvailability.Contains(equipment.Type))
+                {
+                    if (!string.IsNullOrWhiteSpace(r.ProtectionLevel) && protectionLevels.Contains(r.ProtectionLevel.Trim(), StringComparer.OrdinalIgnoreCase))
+                    {
+                        equipment.ProtectionLevel = protectionLevels.Single(a => a.Equals(r.ProtectionLevel.Trim(), StringComparison.OrdinalIgnoreCase));
+                    }
+                    else
+                    {
+                        equipment.ProtectionLevel = "P1";
```
srkirkland: What do you think about defaulting to null, which could be an option in the select list equivalent to "unknown"?  Or we could even make a new status UK which is unknown.  Just want to be careful, not sure if defaulting to P1/A1 is right or not

jSylvestre: @srkirkland With my Script I'm defaulting everything that matches to P1 A1.
If we go with an "unknown" that basically means they can change it to unknown.

jSylvestre: I could have that record error out if you want.

srkirkland: Yea I think changing to unknown might be a good option for people to have.  But for now let's default to P1/A1 and make the selection mandatory and see if that works.  Since it is required for IS3 maybe that's right anyway.  We'll discuss with adam when he comes back too.

Issue #659, opened 2019-07-25
INC0826683 -- Other people import sources


jSylvestre:
User:

I ran into an issue yesterday with the students that I needed to be added.
They are not in PPS and may never be in PPS depending on if they work for
us or not. Every fall I have between 40-70 new MS and PhD students that
need to be added to PEAKS to assign keys and key cards to. The only way I
was able to add them in bulk was to create a mock key and input them with
there kerberos in a CSV document to import. There has to be a better way to
do this for students and visiting scholars who may not ever be in PPS.

jSylvestre:
Perhaps there is a PPS lookup for this?
Another source to bulk load from?
Worst case, create a cvs people import.

jcubbage:
I've tried everything I can to get good student data, and it doesn't work well. IF the students are grad students AND they have either a TA or GSR position, they will be in PPS. But, there are a ton of students in departments that do not fit that. IAM does have student look-up by major, but that may grab many more students then you intend, especially with grad groups (grad students in "nutrition" can be housed in any of 5-6 departments). So, I think CSV import for people is our best bet. This user clearly is using that as a method to get them in by doing key CSV.

PR #673, opened 2019-08-12
Jcs/import people

jSylvestre:
Issue #659

### DIFF
--- a/Keas.Mvc/Controllers/PersonAdminController.cs
+++ b/Keas.Mvc/Controllers/PersonAdminController.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using CsvHelper;
+using CsvHelper.Configuration;
 using Keas.Core.Data;
 using Keas.Core.Domain;
 using Keas.Core.Models;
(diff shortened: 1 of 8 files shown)


Keas.Mvc/Controllers/PersonAdminController.cs
```diff
@@ -159,6 +167,245 @@ await _context.WorkstationAssignments.AnyAsync(a => a.PersonId == person.Id))
...
+                        resultsView.Add(importResult);
+                    }
+                }
+                catch (Exception e)
+                {
+                    counter++;
+                    var importResult = new PeopleImportResult(new PeopleImport());
+                    importResult.Success = false;
+                    importResult.LineNumber = counter;
+                    importResult.ErrorMessage.Add("There is a problem with this row preventing it from being imported. Maybe an invalid date format. Import Halted.");
```
srkirkland: Isn't this exception run once per import and not a row problem?  it doesn't seem to be within the for loop

jcubbage:
Doesn't catch empty/no file upload. The other 2 CSV's return Message = "File not selected". People just errors out.

Keas.Mvc/Controllers/PersonAdminController.cs
jSylvestre: @srkirkland It happens in the actual forEach(var r in records) it happens when a row in records isn't valid. I couldn't figure out a way to trap it per row and actually continue.

PR #677, opened 2019-08-14
Fix IS3 Fields not populating

jSylvestre:
Have a separate method to update the equipment type what will default values.
Otherwise, if the user doesn't change from P1 and A1, then are blank in the db

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/AssignEquipment.tsx
@@ -156,6 +156,7 @@ export default class AssignEquipment extends React.Component<IProps, IState> {
                                                 selectedEquipment={this.state.equipment}
                                                 commonAttributeKeys={this.props.commonAttributeKeys}
                                                 changeProperty={this._changeProperty}
+                                                changeType={this._changeType}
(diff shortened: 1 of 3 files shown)


jSylvestre:
FYI, this is a PR into my other PR and I have not fixed the tests there yet. That is why it is failing.

laholstege:
@srkirkland the problem is that we only want it to default on certain types. if it was a default for all types we could just set it when we create a new equipment. 

@jSylvestre i guess we could also set it in `EquipmentContainer`'s edit and create methods. but what happens if someone changes the type to computer, and then changes to software? shouldn't the defaulted values get removed?

jSylvestre:
@laholstege It isn't really a big deal if there are values there for non-computer types.
If we clear them out, I'd rather do it server side. Which, maybe is the better place to set these as well.

@srkirkland For the other suggested changes, I agree, but I'd say do them in a separate fix.

laholstege:
@jSylvestre if it's not a big deal for non-computer types, why not just set them as default on create? and then if we want we can clean them up later

jSylvestre:
@laholstege The only reason we wouldn't want them set is for reporting. I'm fine with other solutions.

PR #678, opened 2019-08-14
Jcs/bigfix

jSylvestre:


Keas.Core/Domain/Equipment.cs
```diff
@@ -22,7 +22,10 @@ public Equipment()
         public string ProtectionLevel { get; set; }
 
         [StringLength(2)]
-        public string AvailabilityLevel { get; set; }
+        public string AvailabilityLevel { get; set; }
+
+        [StringLength(16)]
+        public string BigfixId { get; set; }
```
srkirkland: You had a good name earlier i don't remember, something like ExternalIdentifier or whatnot.  I don't think we should name the variable itself bigfix but just it's an id to whatever central system comes and goes

Keas.Mvc/ClientApp/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -148,6 +148,23 @@ export default class EquipmentEditValues extends React.Component<IProps, {}> {
...
+                        <div className="form-group">
+                            <label>Bigfix Id</label>
+                            <input
+                                type="text"
+                                className="form-control"
+                                disabled={this.props.disableEditing}
+                                value={
+                                    this.props.selectedEquipment.bigfixId
+                                        ? this.props.selectedEquipment.bigfixId
+                                        : ""
```
srkirkland: Just as an FYI you can simplify this to `this.props.selectedEquipment.bigfixId || ""` if you want

jSylvestre:
Production and Test databases have been updated.

PR #681, opened 2019-08-14
update npm dependencies

srkirkland:
new major versions of webpack & typescript.  updated react + most dependencies, with the exception of react-bootstrap-typeahead & react-router-dom.

Bonus -- should bring back hot reloading for react.


srkirkland:
CSS on non-react pages has to do with webpack bundler, and the serial problem seems to do with the datepicker.  I'm getting "Invalid time value".

PR #682, opened 2019-08-14
add check for null

laholstege:


jSylvestre:
There is a ts error.
Also, do it like this instead?
value={this.props.selectedEquipment.systemManagementId || ""}

Issue #687, opened 2019-08-16
CSV People Import


jSylvestre:
Look into / think about a flag or some other way to "update" existing people. Currently, if the user already exists, the import skips them.

PR #698, opened 2019-08-20
Allow upload people csv to update existing users

jSylvestre:
issue #687

image

### DIFF
--- a/Keas.Mvc/Controllers/PersonAdminController.cs
+++ b/Keas.Mvc/Controllers/PersonAdminController.cs
@@ -169,26 +169,27 @@ namespace Keas.Mvc.Controllers
 
         public IActionResult UploadPeople()
         {
-            var model = new List<PeopleImportResult>();
-            return View();
+            var model = new PeopleImportModel
+            {
+                UpdateExistingUsers = false, ImportResult = new List<PeopleImportResult>()
+            };
+            return View(model);
         }
 
         [HttpPost]
-        public async Task<IActionResult> UploadPeople(IFormFile file, string updateExisting)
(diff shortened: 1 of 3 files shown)


Keas.Mvc/Controllers/PersonAdminController.cs
```diff
@@ -174,8 +174,10 @@ public IActionResult UploadPeople()
         }
 
         [HttpPost]
-        public async Task<IActionResult> UploadPeople(IFormFile file)
+        public async Task<IActionResult> UploadPeople(IFormFile file, string updateExisting)
```
srkirkland: Why not use a bool if you are already using a checkbox?

jSylvestre: Couldn't get it to bind

Keas.Mvc/Views/PersonAdmin/UploadPeople.cshtml
```diff
@@ -15,7 +15,14 @@
                 <form asp-action="UploadPeople" method="post" enctype="multipart/form-data">
                     <div asp-validation-summary="ModelOnly" class="text-danger"></div>
                     <fieldset>
+
                         <div class="form-horizontal">
+                            <div class="form-group">
+                                <div class="row">
+                                <label class="control-label">Update Existing Users Too:</label>
+                                    <input name="UpdateExisting" type="checkbox" class="form-control form-control-lg" value="true"/>
```
srkirkland: try setting checked="checked" and see if that binds.  We should figure out why the bool binding doesn't work first

PR #704, opened 2019-08-21
Lh/update typeahead

laholstege:


Keas.Mvc/package.json
```diff
@@ -14,7 +14,7 @@
     "log4javascript": "1.4.15",
     "moment": "2.24.0",
     "react": "16.9.0",
-    "react-bootstrap-typeahead": "3.2.4",
+    "react-bootstrap-typeahead": "^3.4.6",
```
srkirkland: @laholstege remove the ^ in there and just specify the exact version number.  Too often there is a weird issue otherwise.  Same with the `@types` version

PR #720, opened 2019-08-29
method for querying bigfix computers

srkirkland:


srkirkland:
@jSylvestre FYI this is totally blind, I couldn't run it because of certificate issues with the bigfix endpoint.

PR #721, opened 2019-08-29
Check nulls

jSylvestre:


jSylvestre:
@srkirkland I couldn't reproduce the error, so I just checked it all. For where we are using this in other places, if any of it is null, and exception is probably appropriate. For just the title, it wouldn't matter if we can find it or not

srkirkland:
@jSylvestre in that case i say we just kill this PR if we can reproduce

PR #728, opened 2019-09-03
Nf - bigfix info modal integreted

NabilFurmoli:


Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+        router: PropTypes.object,
+        team: PropTypes.object
+    };
+
+    public state = { bigfixModal: false, computerInfo: {}, isFetched: false };
+
+    public modalToggle = () => {
+        this.setState({
+            bigfixModal: !this.state.bigfixModal
+        });
```
srkirkland: If your new state depends on the old state, you should use the functional form of set state otherwise you can get weird bugs.  ex:
 
```
this.setState(prevState => ({
  showModal: !prevState.showModal
}));
```

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+import * as React from "react";
+import { Button, Modal, ModalBody, ModalFooter, ModalHeader, Table } from "reactstrap";
+
+interface IProps {
+    bigfixId: string;
+}
+
+interface IState {
+    bigfixModal: boolean;
+    computerInfo: object;
```
laholstege: this should be a better type than just `object`, if you need to add an interface to `Types.ts`

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    >
+                        <i className=" ml-2 fas fa-info-circle" />
+                    </a>
+                </label>
+                {this.renderBigFixModal()}
+            </>
+        );
+    }
+
+    private renderBigFixModal = () => {
```
laholstege: not a huge deal, but for consistency i'd name your private methods starting with an underscore, e.g. `_renderBigFixModal()`

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+                method: "GET"
+            }
+        );
+
+        const sortedResult = Object.keys(response)
+            .sort()
+            .reduce((accumulator, currentValue) => {
+                accumulator[currentValue] = response[currentValue];
+                return accumulator;
+            }, {});
```
srkirkland: clever

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export default class EquipmentDetails extends React.Component<IProps, IState> {
+    public static contextTypes = {
+        fetch: PropTypes.func,
+        permissions: PropTypes.array,
+        router: PropTypes.object,
+        team: PropTypes.object
+    };
+
+    public state = { bigfixModal: false, computerInfo: {}, isFetched: false };
```
laholstege: why is this `public state = {...}` and not `this.state = {...}`?

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    public modalToggle = () => {
```
laholstege: does this need to be public?

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,121 @@
...
+    bigfixId: string;
+}
+
+interface IState {
+    bigfixModal: boolean;
+    computerInfo: object;
+    isFetched: boolean;
+}
+
+export default class EquipmentDetails extends React.Component<IProps, IState> {
```
laholstege: this should be `EquipmentBigFixInfo`, not `EquipmentDetails`

laholstege:
@NabilFurmoli  overall looks pretty good! just a few thoughts:
clicking on the info circle when the id is not populated throws an error and is stuck loading
if the bigfix id is wrong, similar problem. i think maybe either we try to do some validation on the bigfix id, or we just have the modal handle the errors better. 
i also think it'd be neat if you can get the button cursor to change when hovering on the icon, so that it's more obvious that it's clickable (like the actual links on that same page)

NabilFurmoli:
@laholstege so for now we if there is no bigfix id, then we don't show the info icon. but as you pointed  we will later still  should consider the cases where the id is invalid.

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: fixed

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: we can declare state outside of the constructor as well. but I am not sure if we should have it as public, private or protected. I will change it to private since it only belongs to the class.

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: fixed

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: yes, it is good convention to initialize state in the constructor. fixed

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: fixed

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
NabilFurmoli: fixed

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -0,0 +1,134 @@
...
+                <label> Bigfix Id</label>
+                <span />
+                {this._renderInfoIcon()}
+                {this._renderBigFixModal()}
+            </>
+        );
+    }
+
+    private _renderInfoIcon = () => {
+        if (this.props.bigfixId) {
```
laholstege: small thing/just an fyi, but i usually try to do:
```
if(!this.props.bigfixId){
  return null;
}
return (...)
```

NabilFurmoli: fixed

Issue #464, opened 2019-01-31
better error handling in react


laholstege:
since we removed generic alert, should have some way of notifying user of error

PR #747, opened 2019-09-11
Lh/workstation toasts

laholstege:
#464

### DIFF
--- a/Keas.Mvc/ClientApp/components/Workstations/RevokeWorkstation.tsx
+++ b/Keas.Mvc/ClientApp/components/Workstations/RevokeWorkstation.tsx
@@ -12,7 +12,7 @@ interface IProps {
 }
 
 interface IState {
-    loading: boolean;
+    submitting: boolean;
 }
 
 export default class RevokeWorkstation extends React.Component<IProps, IState> {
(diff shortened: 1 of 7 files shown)


Keas.Mvc/ClientApp/components/Workstations/RevokeWorkstation.tsx
```diff
@@ -68,7 +75,14 @@ export default class RevokeWorkstation extends React.Component<IProps, IState> {
         if (!this.props.selectedWorkstation) {
             return null;
         }
-        await this.props.revokeWorkstation(this.props.selectedWorkstation);
+        this.setState({ submitting: true });
+        try {
+            await this.props.revokeWorkstation(this.props.selectedWorkstation);
+        } catch (err) {
+            this.setState({ submitting: false });
```
srkirkland: So once you take out the default error log, are you going to come back here and add error toast for each case like this?

srkirkland: or should you add in toast fail/success notifications here?

laholstege: @srkirkland I do the toasts at the actual API call, so in `WorkstationContainer` in this case. then that throws an error that is caught in each modal, so that the modal doesn't close.

srkirkland: ah ok, yea that works too

PR #763, opened 2019-09-16
Lh/final toasts and errors

laholstege:
some small missed stuff + toasts when searching fails 
#764 is the only thing that needs toasts

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
@@ -128,7 +128,11 @@ export default class AccessEditValues extends React.Component<IProps, {}> {
         const accessAssignment = this.props.selectedAccess.assignments.filter(
             x => x.personId === personId
         );
-        // TODO: add submitting state
-        await this.props.onRevoke(accessAssignment[0]);
+        try {
+            await this.props.onRevoke(accessAssignment[0]);
+        } catch (err) {
+            // TODO: add submitting state and handle
+            return;
+        }
     };
 }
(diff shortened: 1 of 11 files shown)


Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
```diff
@@ -128,7 +128,11 @@ export default class AccessEditValues extends React.Component<IProps, {}> {
         const accessAssignment = this.props.selectedAccess.assignments.filter(
             x => x.personId === personId
         );
-        // TODO: add submitting state
-        await this.props.onRevoke(accessAssignment[0]);
+        try {
+            await this.props.onRevoke(accessAssignment[0]);
+        } catch (err) {
+            // TODO: add submitting state and handle
```
jSylvestre: @laholstege Just throw an alert in there for now?

Keas.Mvc/ClientApp/components/Access/SearchAccess.tsx
```diff
@@ -98,6 +90,24 @@ export default class SearchAccess extends React.Component<IProps, IState> {
...
 
+    private _onSearch = async (query: string) => {
+        this.setState({ isSearchLoading: true });
+        let accesses: IAccess[] = null;
+        try {
+            accesses = await this.context.fetch(
+                `/api/${this.context.team.slug}/access/search?q=${query}`
+            );
+        } catch (err) {
+            toast.error("Error searching accesses.");
```
jSylvestre: Don't think this can happen, but good to have.

Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
laholstege: @jSylvestre it'll still have a toast, but eventually i want to clean assets up and this should have a submitting state

PR #767, opened 2019-09-16
Lh/handle deleting for details fetch

### LINKED ISSUE #766: handle fetching on details if asset has been deleted

laholstege:
closes #766

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
@@ -383,8 +383,9 @@ export default class AccessContainer extends React.Component<IProps, IState> {
         }
     };
 
-    private _updateAccessFromDetails = (access: IAccess) => {
-        const index = this.state.accesses.findIndex(x => x.id === access.id);
+    private _updateAccessFromDetails = (access: IAccess, id?: number) => {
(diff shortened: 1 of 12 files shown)


Keas.Mvc/ClientApp/components/Access/AccessContainer.tsx
```diff
@@ -383,8 +383,9 @@ export default class AccessContainer extends React.Component<IProps, IState> {
         }
     };
 
-    private _updateAccessFromDetails = (access: IAccess) => {
-        const index = this.state.accesses.findIndex(x => x.id === access.id);
+    private _updateAccessFromDetails = (access: IAccess, id?: number) => {
```
srkirkland: Why this strategy of adding a new id param instead of just doing something like `const id = access?.id`?  Seems like it would tell you the same info

laholstege: @srkirkland it's because if what they're trying to look at has been deleted, `access` will be null. but i still want the id so i can update the state, so i send that separately

srkirkland: ooh so it's not the access.id really, it's the old deleted accessId that shows up there.  ok 🚢

PR #768, opened 2019-09-16
item -> name, code -> key code, number -> key serial number

### LINKED ISSUE #730: Consistent Titles
It would be helpful if the titles were consistent when inputting keys into PEAKS (ie Key Name, Key Code, Key Serial).  When creating a key the key name is titled: Item.  When adding a key serial the key serial is titled: number.  Thank you! 

--------------------
Votes: 2
Author: sjoseph
Voters: sjoseph2,rcasteli

laholstege:
closes #730

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
@@ -71,7 +71,7 @@ export default class AccessEditValues extends React.Component<IProps, {}> {
                 )}
                 <div className="wrapperasset">
                     <FormGroup>
-                        <Label for="item">Item</Label>
+                        <Label for="name">Name</Label>
                         <Input
                             type="text"
                             className="form-control"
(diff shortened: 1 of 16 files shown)


jSylvestre:
@laholstege If you close a feedback issue, it hides it from the feedback site.
It would be nice for you to contact the people listed. You can use who.ucdavis.edu to get their emails.
Or if you don't want to, just slack me the issue and I'll do it.

Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
```diff
@@ -71,7 +71,7 @@ export default class AccessEditValues extends React.Component<IProps, {}> {
                 )}
                 <div className="wrapperasset">
                     <FormGroup>
-                        <Label for="item">Item</Label>
+                        <Label for="name">Name</Label>
```
srkirkland: the `for="name"` doesn't actually do anything unless we added a name attr to the related input, which maybe we should do.  I think it's an accessibility thing.  Nothing to do with your PR really, just something that stuck out

PR #772, opened 2019-09-18
Nf/updated bigfix components and fetch errors handling

NabilFurmoli:


Keas.Mvc/Services/BigfixService.cs
```diff
@@ -66,12 +66,19 @@ public async Task<BigfixComputerSearchResult[]> GetComputersByName(string name)
...
-            using (var bf = GetClient())
+            try {
+                using (var bf = GetClient())
+                {
+                    var results = await bf.Computers.Get(id);
+
+                    return new BigFixComputerProperties(results);
+                }
+            } 
+            catch (Exception)
```
srkirkland: Do no swallow exceptions inside of a service.  Other callers may also use this service and should have the option of catching specific errors

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -324,16 +324,24 @@ public async Task<IActionResult> GetHistory(int id)
             return Json(history);
         }
 
-        public async Task<BigFixComputerProperties> GetComputer(string id)
+        public async Task<IActionResult> GetComputer(string id)
         {
-            return await this._bigfixService.GetComputer(id);
+            var result = await this._bigfixService.GetComputer(id);
```
srkirkland: if you want to try/catch here, that would be a better spot.  Also, why not just handle it in the client?  is the appropriate data not being transferred through?

Keas.Mvc/Services/BigfixService.cs
NabilFurmoli: fixed

Keas.Mvc/Controllers/Api/EquipmentController.cs
NabilFurmoli: so, as you mentioned, for now, I handled the exception here. one thing I am not quite sure is that how bigfix server handles the invalid id. if we don't catch the exception here, on our client-side we receive a 500 Internal Server Error.

NabilFurmoli:
@jSylvestre the PR does not add new features to the application, in this PR I have only reconstructed Bigfix related components to make them more readable and cleaner, also I have updated error handling logics used in those components for fetching data.

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -127,27 +110,34 @@ export default class EquipmentBigFixInfo extends React.Component<IProps, IState>
...
-        if (!response.ok) {
-            // show the invalid Bigfix-id message.
+        let response = null;
+        try {
+            response = await this.context.fetch(
+                `/api/${this.context.team.slug}/equipment/GetComputer/${id}`
+            );
+        } catch (err) {
+            if (err.message === "Not Found") {
+                toast.error("Not a valid Bigfix id, please make sure to enter a valid Bigfix id.");
```
laholstege: @NabilFurmoli this shouldn't be a toast, the modal should pop up with an error telling the user that their id is invalid.

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixSearchId.tsx
```diff
@@ -13,13 +16,16 @@ import {
 } from 'reactstrap';
 import { Context } from '../../Context';
 import { IBigFixSearchedName } from '../../Types';
+import { threadId } from 'worker_threads';
+import { isThisQuarter } from 'date-fns';
```
srkirkland: @NabilFurmoli remove these lines, they are not used anywhere

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -324,16 +325,41 @@ public async Task<IActionResult> GetHistory(int id)
...
+                 return Json(result);
+            }
+            catch (BigfixApiException ex) 
+            {
+                if (ex.StatusCode == System.Net.HttpStatusCode.NotFound)
+                {
+                    return NotFound();
+                }
+                else
+                    throw;
```
srkirkland: keep using `{ }` for else statements even if they are one-liners

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -324,16 +325,41 @@ public async Task<IActionResult> GetHistory(int id)
...
+                {
+                    var result = await this._bigfixService.GetComputersByName(value);
+                    if (result.Length == 0)
+                        return NotFound();
+
+                    else
+                        return Json(result);
+
+                } catch (BigfixApiException) {
+                    throw;
```
srkirkland: @NabilFurmoli don't bother catching if you are just going to re-throw

Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixSearchId.tsx
```diff
@@ -219,45 +233,61 @@ export default class EquipmentBigFixSearchId extends React.Component<
...
-        isValidSearch: false
-      });
-    } else {
-      const firstFiveName = response.slice(0, 5);
-      this.setState({
-        isFetched: true,
-        isSearching: false,
-        listOfComputers: firstFiveName
-      });
+      debugger;
```
laholstege: @NabilFurmoli you should take this out, and fix the other tslint warnings on this page. but other than that looks good! :shipit:

PR #773, opened 2019-09-19
Update StringExtensions.cs

NabilFurmoli:
on the server-side, if formating a phone number throws an exception we simply return the value.

### DIFF
--- a/Keas.Core/Extensions/StringExtensions.cs
+++ b/Keas.Core/Extensions/StringExtensions.cs
@@ -4,22 +4,29 @@ using System.Text;
 using Humanizer;
 using PhoneNumbers;
 
-namespace Keas.Core.Extensions {
-    public static class StringExtensions {
-        public static string SafeHumanizeTitle (this string value) {
-            if (string.IsNullOrWhiteSpace (value)) {
+namespace Keas.Core.Extensions
+{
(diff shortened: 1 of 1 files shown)


jSylvestre:
I really don't like the style of putting the { on the same line.

NabilFurmoli:
I fixed the code format, I was using a weird C# format extension in vscode apparently.

Keas.Core/Extensions/StringExtensions.cs
```diff
@@ -23,19 +23,26 @@ public static string FormatPhone(this string value)
...
+            var phone = "";
+            try 
             {
-                if (phoneNumberUtil.IsValidNumber(phoneNumber))
-                {
-                    return phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.INTERNATIONAL);
-                }
+                var phoneNumberUtil = PhoneNumbers.PhoneNumberUtil.GetInstance ();
+                var phoneNumber = phoneNumberUtil.ParseAndKeepRawInput (value, "US");
+                phone = phoneNumberUtil.Format (phoneNumber, PhoneNumberFormat.NATIONAL);
```
srkirkland: This doesn't look used anywhere

PR #785, opened 2019-09-26
Lh/autofocus person

### LINKED ISSUE #784: autofocus assign person input
possibly move "Go!" Button so user doesn't have to scroll to the very bottom

laholstege:
autofocuses person input when opening assign modal
closes #784


laholstege:
weird interaction with modals if we disable autofocus on the modal

Issue #736, opened 2019-09-05
update context, react-router


laholstege:
see if react-router-dom 5.0 works well enough to use

laholstege:
https://github.com/ReactTraining/react-router/issues/6885

PR #791, opened 2019-10-01
Lh/react router

laholstege:
#736


laholstege:
there's weird interactions between spaces and keys that i still need to resolve

PR #799, opened 2019-10-08
Lh/validation

laholstege:
validation for key serials. if we like this pattern i'll add it to everything else


Keas.Mvc/ClientApp/components/Keys/EditKeySerial.tsx
```diff
@@ -146,12 +151,20 @@ export default class EditKeySerial extends React.Component<IProps, IState> {
...
+      error = 'You must give this key serial a name.';
     } else if (this.state.keySerial.number.length > 64) {
       valid = false;
-      error = 'The name you have chosen is too long';
+      error = 'The serial number you have chosen is too long';
+    } else if (
+      !this.props.checkIfKeySerialNumberIsValid(
+        this.state.keySerial.number,
+        this.state.keySerial.id
+      )
```
srkirkland: semi colon?

PR #809, opened 2019-10-12
Peaks 748 access

### LINKED ISSUE #748: make access more consistent
- [x] clean up SearchAccess, only search by name
image
- [x] move Access Assignment table out of `AccessEditValues.tsx` and into its own component
- [x] move details page out of modal and make separate page (like keys or people details)
- [ ] add revoke modal
- [x] put "add access" button on details page

klanmiko:
Closes #748 

This ended up being a larger set of changes than I anticipated.

The smaller set of changes involved moving the assignment table out into its own view.

However, since we also want to show a modal for when an assignment is revoked. I needed to create a separate route for one the revoke modal is shown, and I encapsulated revoking an access into its own assignmentContainer.

I had to swap out the access container from the people page with the assignment container, because it is hard for the access container to figure out, based on props and route, whether it is a root container. This also prevents the user from deleting an entire access from within the people page, and follows more closely the model of keyserials

I would be able to simplify this PR by keeping the AssignmentContainer changes within the Access Container, though I am not sure if this would be better

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AssignmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AssignmentContainer.tsx
@@ -36,9 +36,10 @@ class AssignmentContainer extends React.Component<IProps, IState> {
     super(props);
 
     const assignments = (props.access && props.access.assignments) || [];
+    const {action} = props.match.params
 
     this.state = {
-      isRevokeModalShown: props.match.params.action === 'revoke',
+      isRevokeModalShown: action === 'revoke',
       assignments: assignments,
       selectedAssignment: assignments.find(
         el => el.id === parseInt(props.match.params.id, 10)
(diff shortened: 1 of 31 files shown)


Keas.Mvc/ClientApp/components/Access/AssignmentContainer.tsx
```diff
@@ -0,0 +1,263 @@
...
+    );
+  };
+
+  private hideModals = () => {
+    this.setState({
+      isRevokeModalShown: false,
+      isAssignModalShown: false,
+      selectedAssignment: null
+    });
+    this.props.history.replace(this._getBaseUrl());
```
laholstege: is there a reason you're doing replace instead of push here?

klanmiko: My reasoning was that they shouldn't be able to use the back button to bring up the modal again, since it has already been revoked.

klanmiko:
1. I formatted some files but I'll format all the changed files.

2. I'll look

3. Yea I can do that, are there any more UI related things that could be better with the assignment table and details page?

4. Can I do that as a separate PR? I feel like this is too big

laholstege:
@klanmiko 
probably remove the modal styles (e.g. the green side panel, new component to render the details that isn't using the `AccessEditValues`) but we can also do that in another PR / with cal's approval. i'm fine with you adding other stuff in another PR

Keas.Mvc/ClientApp/components/Access/RevokeAccess.tsx
```diff
@@ -0,0 +1,68 @@
...
+            color='primary'
+            onClick={() => {
+              this.setState({submitting: true})
+              this.props.revoke(this.props.assignment).catch(e => {
+                this.setState({
+                  error: "Error -- " + e.message + " -- Please try again later",
+                  submitting: false,
+                })
+              })
+            }}
```
laholstege: @klanmiko i'd prefer this be its own function instead, and following the same pattern as the other revokes

jSylvestre:
from people page, equipment revoke:

image

klanmiko:
@jSylvestre Fixed your revoke issue, though I may have broken editing an access, though we should also test that @laholstege.

PR #811, opened 2019-10-17
Jcs/email in report20191017

jSylvestre:


Keas.Mvc/Views/Report/UnAcceptedItems.cshtml
```diff
@@ -4,14 +4,6 @@
     ViewData["Title"] = "Items Pending Acceptance";
 }
 
-@section Styles {
-    @await Html.PartialAsync("_ReportStylePartial")
-}
-
-@section Scripts {
-   @await Html.PartialAsync("_ReportScriptsPartial")
```
srkirkland: Instead of custom doing the report scripts, why not just make all of the reports exportable?

PR #814, opened 2019-10-22
only build PRs and master branch

srkirkland:


srkirkland:
won't fix the problem but might make our builds more efficient by not building branches we don't care about

PR #816, opened 2019-10-23
yup for key serials

laholstege:
adds schemas for keys/key serials. uses schemas for validation for key serials in `AssignKeySerial` and `EditKeySerial`. also I deleted `CreateKeySerial` because we never use it. I'll add validation for the rest of keys in my next PR


jSylvestre:
@laholstege 

image

PR #821, opened 2019-10-29
Km/peaks 812 access

### LINKED ISSUE #812: Access Assignments improvements 2: Electric Boogaloo
- [x] move the expiration date to be right below the assign person when assigning an access

- [ ] Allow revoke from anywhere the assignment table is present, except on assignment page

- [ ] improve styles on revoke modal

- [x] new component to render the details that isn't using the AccessEditValues that does not have the green sidebar

- [x] Add a delete button to the access details page

- [x] Assign button on access table should route to the details page

klanmiko:
Fix #812 

Refactor to move the Access Assignment table out of AccessEditValues. AccessEditValues can now take an argument children to display with the left green border.

Access Assignment Card has been refactored out so that both the list and table can be composed with it.


srkirkland:
@jSylvestre can you look at this too along with @laholstege .  I thought the code all looked good but just to check that it doesn't involve any bugs.  Basically just double check that the access functions work.

klanmiko:
@jSylvestre I fixed the regression

klanmiko:
The issue was that the modal was not being closed. This has been fixed

jSylvestre:
From the person page, assigning a new access does not update the page correctly:

image


Refreshing the page updates the info

srkirkland:
@klanmiko looks like something has changed that needs some conflict resolution.  Please fix & then we'll test and merge it.

PR #822, opened 2019-10-30
handle bigfix 403

laholstege:
not sure if we want to display it as a toast, but here's how to do it @NabilFurmoli

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
@@ -136,6 +136,14 @@ export default class EquipmentBigFixInfo extends React.Component<
           isFetched: true,
           isFound: false
         });
+      } else if (err.message === 'Forbidden') {
+        this.setState({
+          isFetched: true,
+          isValidRequest: false
+        });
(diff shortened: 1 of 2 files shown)


Keas.Mvc/ClientApp/components/Equipment/EquipmentBigFixInfo.tsx
```diff
@@ -136,6 +136,14 @@ export default class EquipmentBigFixInfo extends React.Component<
           isFetched: true,
           isFound: false
         });
+      } else if (err.message === 'Forbidden') {
```
srkirkland: @laholstege We're actually returning the proper 403 status code here -- is reading the error message the best we can do?  Is the fetch library hiding the status code on the client side somehow?

laholstege: @srkirkland our fetch in `api.ts` is swallowing it and only throwing `res.statusText`. i'm happy to change it, that's just how it is now

srkirkland: @laholstege oh i see that now, i'll need to have words with the guy who wrote that.

Issue #805, opened 2019-10-10
better validation on react assign/edit


laholstege:
- [x] keys
- [x] key serials
- [x] equipment
- [ ] access
@NabilFurmoli do these:
- [ ] workstations
- [ ] people

PR #825, opened 2019-10-31
Lh/yup keys

laholstege:
#805 
#828 

adds yup validation for keys and key serials, errors are displayed under the input the error occurred in


Keas.Mvc/ClientApp/components/Keys/EditKey.tsx
```diff
@@ -14,7 +15,8 @@ interface IProps {
 }
 
 interface IState {
-  error: string;
+  errorMessage: string;
+  errorPath: string;
```
srkirkland: @laholstege what do you think about combining this into like a `IValidationError` object or something, that way you can keep the state clean and only pass one object, and plus it also leaves room for other error types.  You might even just be able to use the yup native `ValidationError` object too, or something similar to it.

PR #830, opened 2019-11-07
Lh/Validate date properly 

laholstege:
use generalized AssignDate to render FormGroup and Label, make AssignPerson follow same pattern
#805


jSylvestre:
@laholstege See ok when updating, but not adding a new serial and assigning then?

image

PR #848, opened 2019-11-21
Use config settings to create super user

jSylvestre:


jSylvestre:
@srkirkland Once this is deployed, we can remove any super access from prod without it getting added back.

jSylvestre:
> @jSylvestre won't this only allow a single super user though?

@srkirkland No, it adds the superuser if it isn't there. So locally, you could just add yourself (UserSecret).
If we needed to add more than one user in the future, we could just stop-start after changing it.

Issue #828, opened 2019-11-05
use readonly not disabled on editvalues 


laholstege:
- [x] keys
- [x] key serials
- [x] equipment
- [ ] access
@NabilFurmoli do these:
- [ ] workstations
- [ ] people

PR #854, opened 2019-11-26
Lh/access schema

laholstege:
#828 #805


laholstege:
merging #855  will merge this

PR #863, opened 2020-01-15
look at keyId when checking if serial number is valid

### LINKED ISSUE #861: Add Key Serial from people page
Go isn't enabled, but it is from a different person pr from the key tab

laholstege:
closes #861


jSylvestre:
Fixes #861

PR #864, opened 2020-01-16
Try to find it without splitting the terms first

jSylvestre:
Not sure if I like this...
By using the tetms.any(...) it will "or" them together. 

It only calls the second more permissive search if 5 or less results are returned.

Also... This appears to be called after every keystroke...


jSylvestre:
Killing PR. Too old.

PR #884, opened 2020-07-06
Jcs/title and stuff

### LINKED ISSUE #857: Get Title
From IAM should probably not throw an exception because we probably don't care that much...

jSylvestre:
Closes #857

### DIFF
--- a/Keas.Mvc/Services/IdentityService.cs
+++ b/Keas.Mvc/Services/IdentityService.cs
@@ -259,7 +259,7 @@ namespace Keas.Mvc.Services
             }
             catch (Exception ex)
             {
-                Log.Error($"Getting Title for IamId {iamId}. Exception: {ex.Message} -- {ex.InnerException.Message} ");
+                Log.Error($"Getting Title for IamId: {iamId}.", ex);
             }
 
             return title;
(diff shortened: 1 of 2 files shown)


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -247,12 +248,18 @@ public async Task<PPSDepartmentResult> GetPpsDepartment(string ppsCode)
...
+                var clientws = new IetClient(_authSettings.IamKey);
+                var result = await clientws.PPSAssociations.Search(PPSAssociationsSearchField.iamId, iamId);
+                if (result.ResponseData.Results.Length > 0)
+                {
+                    title = result.ResponseData.Results.FirstOrDefault(a => !string.IsNullOrWhiteSpace(a.titleOfficialName))?.titleOfficialName;
+                }
+            }
+            catch (Exception ex)
+            {
+                Log.Error($"Getting Title for IamId {iamId}. Exception: {ex.Message} -- {ex.InnerException.Message} ");
```
srkirkland, line 262: @jSylvestre isn't there a version of `Log.Error` which takes in the actual error instead of just the message?  That way you get a stack trace and other more useful info than writing the exception out into a string.

PR #885, opened 2020-07-06
Tags in Context

### LINKED ISSUE #817: put tags and filters into context

wongband:
- Had asset controller return tags
- Added tags into React Context
- Replaced all instances of `this.state.tags` with `this.context.tags` and removed tags from the state
- Removed api calls to `api/team/tags/listTags`
closes #817


wongband:
Fixed above comment ^^

PR #889, opened 2020-07-07
Updated Js Libraries

### LINKED ISSUE #886: Update client libraries
Update all of the JS libs (in MVC project package.json) to their latest versions.  I'd start with updating react and related libraries first, then move on to smaller libs and finally webpack related libs.  If you run into a problem with any libs, just cancel the update and we'll deal with it individually.  My guess is you are most likely to have problems with typescript files, type definition files, the hotloader, and webpack related build stuff.

As things are updated, periodically check that they still work by going to the assets page and clicking around, and you can also manually run the webpack script and make sure the output doesn't give you any errors.

If you run into issues, here's two fully-updated (in the last month) repos which can be serve as an example

https://github.com/ucdavis/Sloth/blob/master/Sloth.Web/package.json
https://github.com/ucdavis/Namster/blob/master/src/Namster/package.json

wongband:
- Updated all js packages that needed updating
- Removed all jasmine and karma related packages
closes #886


wongband:
Okie ill change it rn

PR #893, opened 2020-07-09
WIP React Table v7 update

srkirkland:
Updating to the new hooks based React Tables v7.  Created wrapper around hooks table component called `ReactTable` which takes in columns + data.  All existing tables are broken except the People table has been fixed up to work with the new code (but needs styling help).  Most tables will be easy to convert with the exception of custom header filters (like the expiration dropdown, for which you should take a look at https://github.com/tannerlinsley/react-table/blob/master/docs/examples/simple.md for an example, and also in our existing Filtering.tsx file).


Keas.Mvc/ClientApp/components/People/PeopleTable.tsx
```diff
@@ -68,7 +86,7 @@ export default class PeopleTable extends React.Component<IProps, {}> {
 
     const initialState: Partial<TableState<any>> = {
       sortBy: [{ id: 'name' }],
-      pageSize: ReactTableUtil.getPageSize(),
+      pageSize: ReactTableUtil.getPageSize()
```
jSylvestre, line 89: Didn't we decide to go with trailing commas?

srkirkland: If we did, we didn't codify it in the peaks config file

srkirkland:
Added filtering, so the remaining tables should be pretty easy to fix up with the exception of the custom column filters (like for serials or expiration date). This example shows how to do it https://codesandbox.io/embed/github/tannerlinsley/react-table/tree/master/examples/filtering

srkirkland:
closed in favor of #899

PR #899, opened 2020-07-14
Updated React Table to v7

wongband:
- Updated the following tables:
  - KeyTable
  - EquipmentTable
  - SpaceTable
  - AccessTable


Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
```diff
@@ -18,6 +17,72 @@ interface IProps {
...
+            onClick={() => this.props.showDetails(data.row.original)}
+          >
+            Details
+          </Button>
+        ),
+        Header: 'Access',
+        filterable: false,
+        maxWidth: 150,
+        resizable: false,
+        sortable: false
```
srkirkland, line 34: I don't think`sortable` is a valid column option.  Neither for that matter is `filterMethod` or `sortMethod` or several of the other ones.  If you use typescript typings you'll be able to see pretty easily, also the api def is here https://react-table.tanstack.com/docs/api/useFilters#column-options but a little confusing because the sort, filter, and general options are in different places

srkirkland: And by use typescript I mean doing `const columns: Column<IAccess>[]` kind of thing.  That'll make sure you aren't using properties that aren't doing anything

Keas.Mvc/ClientApp/components/Keys/KeyTable.tsx
```diff
@@ -27,6 +26,134 @@ interface IRow {
...
+        accessor: 'key.name',
+        className: 'word-wrap',
+        filterMethod: (filter: IFilter, row: IRow) =>
+          !!row[filter.id] &&
+          row[filter.id].toLowerCase().includes(filter.value.toLowerCase())
+      },
+      {
+        Header: 'Key Code',
+        accessor: 'key.code',
+        filterMethod: (filter: IFilter, row: IRow) =>
```
srkirkland, line 58: I have a feeling none of these are doing anything, but it's worth checking.  I don't think any of this code is being called though, so compare the old version on test to your new version to make sure the filtering is the same.

Keas.Mvc/ClientApp/components/Keys/KeySerialTable.tsx
```diff
@@ -16,142 +20,184 @@ interface IProps {
...
+  // Render a multi-select box
+  return (
+    <select
+      value={filterValue}
+      style={{ width: '100%' }}
+      onChange={e => {
+        setFilter(e.target.value || undefined);
+      }}
+    >
+      {ReactTableStatusUtil.defaultFilterOptions.map(option => (
```
srkirkland, line 41: @wongband feels here like the ReactTableStatusUtil is a bit overkill and it would be easier to read if you just wrote out the `<option></option>` in the select.  It would be a lot less lines

Also it looks like you are duplicating the filter.  I see two different `<select>`, here and in the static class.  And also two different filters with basically the same logic.  Did you just keep around the old stuff that isn't used anymore?

Keas.Mvc/ClientApp/components/Keys/KeySerialTable.tsx
```diff
@@ -16,142 +20,184 @@ interface IProps {
...
+  }
+  if (filterValue === 'special') {
+    return rows.filter(r => getRowStatus(r) === 'Special');
+  }
+  return rows;
+};
+
+const getRowStatus = (row: any) => row.original?.status;
+
+class ReactTableStatusUtil {
```
srkirkland, line 71: I think you can dump everything in here.  same with the other places you did the same, like serials and such

PR #902, opened 2020-07-16
Transition away from Sparkpost library to raw SMTP

### LINKED ISSUE #874: Rewrite email to use SMTP
We want to remove the SparkPost lib dependency and instead send using basic SMTP (using the sparkpost server).  Microsoft recommends using https://github.com/jstedfast/MailKit#using-mailkit, which seems like a good choice.

I like to use mailtrap.io as a fake SMTP server to test with (login info in 1pass).

Should be a very straightforward move but it will require updating configuration settings in the appconfig and secrets files, as well as during deployments.  Make sure to retain the same email features, like HTML & TXT versions.

srkirkland:
SparkPost Lib is not compatible with .net standard and causes other problems and we don't need it.  This will require adding some smtp settings to production before we deploy.  The emails themselves and routing should be exactly the same, as we are still using SparkPost for the actual sending, just via another method.

closes #874

### DIFF
--- a/Keas.Core/Keas.Core.csproj
+++ b/Keas.Core/Keas.Core.csproj
@@ -9,6 +9,5 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.0.0" />
     <PackageReference Include="RazorLight" Version="2.0.0-beta1" />
     <PackageReference Include="Serilog" Version="2.7.1" />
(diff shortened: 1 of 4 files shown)


srkirkland:
@jSylvestre FYI I added the correct settings to production so we can deploy when ready.  Maybe schedule for as soon as I get back towards the end of July.

Issue #715, opened 2019-08-28
Write API


UCDFeedbackBot:
It would be great to have write API as it would automate a lot of things currently. 

Some ideas that having write API access would allow:

1. I have automated reporting on when people leave the university so it could set an end date for certain people letting me (or the key admins) know which keys need to be collected.
2. I could keep people in sync by checking their department code. I can also edit their category or tags (such as if they're a TA or not) by pulling from data from my staff. I have data from Grad Hub of students and their researchers and it would need updating whenever a grad student changes advisors (and I could have PEAKS automatically reflect that).
3. For keycards, they're constantly in flux of which rooms people get access to and when I assign and remove keycards. Being able to export that data from Millenium and automatically push it to PEAKS would be great. 
4. If I could write associations for keys as well, it would save me a lot of manual work too since keys/keycards often unlock many rooms.

tl;dr: having an API to edit people and add/remove keys and their space associations = great

--------------------
Votes: 5
Author: rcasteli
Voters: rcasteli,shchee,guatemal,viendih,rfoster

srkirkland:
@sprucely or @jSylvestre could you gather the emails of the voters on this issue and tell them the write API is ready to go, then close the issue?  thanks!

sprucely:
Will do...

jSylvestre:
crbarahona@ucdavis.edu
rtcastelino@ucdavis.edu
rfoster@ucdavis.edu
shchee@ucdavis.edu
vnhoang@ucdavis.edu

PR #905, opened 2020-07-16
Reconfigure Access to allow API access to certain controllers

srkirkland:
Added a new policy requirement `VerifyRoleOrAuthToken` which will grant access to a given team based on a passed auth token header.  If the header isn't available, it just reverts back to the previous role checks (CAS + Security Service).

To make things easier, I made the `SuperController` no longer doing a default `[Authorize]` and instead we allow the controllers themselves to decide.  Most controllers do already.  We could also do this by using an override attribute or a different base controller, but this seemed the best from a flexibility perspective.

This basically gives us a write API ala #715, though we'll still want to document it with swagger and possibly add some convenience methods.  But this guarantees that the API can do everything since it's what we are using to make the asset page itself.

### DIFF
--- a/Keas.Mvc/Handlers/ApiKeyMiddleware.cs
+++ b/Keas.Mvc/Handlers/ApiKeyMiddleware.cs
@@ -4,8 +4,10 @@ using System.Threading.Tasks;
 using Keas.Core.Data;
 using Keas.Core.Domain;
 using Keas.Mvc.Helpers;
+using Keas.Mvc.Models;
 using Microsoft.AspNetCore.Http;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Options;
 
 namespace Keas.Mvc.Handlers
 {
@@ -19,8 +21,10 @@ namespace Keas.Mvc.Handlers
         {
             _next = next;
         }
-        public Task Invoke(HttpContext context, ApplicationDbContext dbContext)
+        public Task Invoke(HttpContext context, ApplicationDbContext dbContext, IOptions<ApiSettings> apiSettingsOptions)
         {
+            var apiSettings = apiSettingsOptions.Value;
+
             // check for header
             if (!context.Request.Headers.ContainsKey(HeaderKey))
             {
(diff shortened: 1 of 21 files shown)


Keas.Mvc/Handlers/VerifyAuthTokenHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var tempData = _tempDataDictionaryFactory.GetTempData(_httpContext.HttpContext);
+                team = Convert.ToString(tempData["TeamName"]);
+            }
+
+            // user has access to team based on claims
+
+            if (context.User.HasClaim("APIKEYTEAM", team)) {
+                context.Succeed(requirement);
+            }
```
sprucely, line 46: Is this supposed to fall through, or should there be a return?

srkirkland: @sprucely good catch -- I think the fallthrough is functionally fine, but it would be better to just quick return once we get success.  I'll change it

Keas.Mvc/Handlers/ApiKeyMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            // check for header
+            if (!context.Request.Headers.ContainsKey(HeaderKey))
+            {
+                return _next(context);
+            }
+            var headerValue = context.Request.Headers[HeaderKey].FirstOrDefault();
+
+            // lookup team that this API key has acess to
+            var teamAccess = dbContext.Teams.FirstOrDefault(t => t.ApiCode == new Guid(headerValue));
```
jSylvestre, line 28: @srkirkland I think we should also create a flag that says Ok to write. Either that, or create a new ApiCode for writing.

jSylvestre:
We also need to deal with all the history/email actions that record who did something.

srkirkland:
> We also need to deal with all the history/email actions that record who did something.

@jSylvestre yes that's true, and an interesting one.  We use the security service to return a `Person` object for tracking, so the best thing to do would be to keep that pattern.  Which might mean creating a "APIUser" Person in the db and returning that person from the security service's `GetPerson()` method.  Currently we never actually use the stored ActorId property so it doesn't even need to be a real db user if we don't want.

PR #908, opened 2020-07-22
Implemented Jest tests for more containers

wongband:
Created jest tests for the following containers:
- EquipmentContainer
- AccessContainer
- SpacesContainer
- KeyContainer


Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
```diff
@@ -28,7 +28,7 @@ export default class AccessTable extends React.Component<IProps, {}> {
             Details
           </Button>
         ),
-        Header: 'Access',
+        Header: 'Details',
```
jSylvestre, line 31: I don't know that you should have a column heading for the Details column. Looks a little funny.
Previous look:
image

New Look:
image

Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
```diff
@@ -63,7 +63,6 @@ export default class AccessTable extends React.Component<IProps, {}> {
     ];
     
     const initialState: Partial<TableState<any>> = {
-      sortBy: [{ id: 'name' }],
```
jSylvestre, line 66: Does it not have an initial sort?

Keas.Mvc/ClientApp/components/Equipment/EquipmentTable.tsx
```diff
@@ -32,7 +32,7 @@ export default class EquipmentTable extends React.Component<IProps, {}> {
             Details
           </Button>
         ),
-        Header: 'Equipment Actions',
+        Header: 'Details',
```
jSylvestre, line 35: Same as above. If possible, I think this would look better without a header.

Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
wongband: If the Header is empty then the application will break. Not sure how to do it without adding something there

Keas.Mvc/ClientApp/components/Equipment/EquipmentTable.tsx
wongband: Same as previous comment, without the a header, the application will break

Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
wongband: Nope

Keas.Mvc/ClientApp/components/Access/AccessTable.tsx
jSylvestre: @wongband How about an space?
Header: ' ',

wongband: Wait actually, I was being silly putting ' ' as the header will work. Should I do that for all detail columns?

jSylvestre: Yes, I think that looks better. Keep the Actions header though.

wongband: Just updated all columns

PR #909, opened 2020-07-22
Made Date-Picker disabled on deleting person modal

### LINKED ISSUE #907: Start/end date calendar on person delete should be read-only
Start/end date calendar on person delete should be read-only (as is, it breaks the site if the values is changed).  It should only be readonly when the modal is deleting, not editing.

ex: http://peaks-test.ucdavis.edu/caes-cru/people/details/4578.

wongband:
Added a new prop for PersonEditValues component, `isDeleting`
If false then the the user can edit the date
If true then the field will be disabled

closes #907


wongband:
@jSylvestre when editing the start or end date it doesn't clear out the screen

PR #921, opened 2020-07-30
Update to AspNetCore 3.1.5 and add SwaggerGen to api endpoints

sprucely:


Keas.Mvc/ClientApp/pages/assets/AssetNav.tsx
```diff
@@ -14,7 +14,7 @@ export default class AssetNav extends React.Component<{}, {}> {
               className='tab-link tab-people'
               activeClassName='active'
             >
-              People
+              Peoples
```
srkirkland, line 17: @sprucely Do we really want to change this? I guess we should ask Cal but we've always used People as the plural in this app, not peoples

Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -34,13 +35,11 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
             // get team name from url
             var team = "";
-            if (context.Resource is AuthorizationFilterContext mvcContext)
+            if (context.Resource is Endpoint)
             {
-                if (mvcContext.RouteData.Values["teamName"] != null)
-                {
-                    team = mvcContext.RouteData.Values["teamName"].ToString();
-                }
+                team = _httpContext.HttpContext.Request.RouteValues["teamName"].ToString();
```
srkirkland, line 40: @sprucely so at this point are we guaranteed that teamName is never null and we don't have to check anymore?

Keas.Mvc/Startup.cs
```diff
@@ -208,51 +272,76 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
             }
 
+
             app.UseStatusCodePages("text/plain", "Status code page, status code: {0}");
 
-            app.UseStaticFiles();
+            if (env.IsDevelopment())
+            {
+                // dist folder will be served by webpack-dev-server
+                app.UseWhen(context => !context.Request.Path.StartsWithSegments("/dist"),
```
srkirkland, line 281: Why is this needed? I've never seen it in any of our other apps.

Keas.Mvc/package.json
```diff
@@ -5,9 +5,11 @@
   "scripts": {
     "debug": "ASPNETCORE_ENVIRONMENT=Development dotnet watch run",
     "webpack": "node node_modules/webpack/bin/webpack.js --env.dev --watch",
+    "devpack": "node node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.config.js",
```
srkirkland, line 8: we can do this in a separate PR, but for other apps i've included the `--hot` param on here in order to get hot reloading working.  it probably requires a bit of other config too but might explain part of why hot reloading isn't working yet

Keas.Mvc/ClientApp/pages/assets/AssetNav.tsx
sprucely: Not something I would deliberately do. I wonder how it happened,

Keas.Mvc/Handlers/VerifyRoleAccessHandler.cs
sprucely: Never saw a null during debugging, but lots of empty strings. I'll add a null coalesce just to be safe...

Keas.Mvc/Startup.cs
sprucely: If the static file middleware picks up the vendor.js and other files, it doesn't give the webpack-dev-server a jance to serve them and inject the HMR code. Took me forever to figure it out, because I thought it was something to do with having just added SwaggerGen.

Keas.Mvc/package.json
sprucely: As far as I'm aware, it's equivalent to the hot: true setting I put inside the webpack.config.

sprucely: I tried it both ways, and it didn't make any difference.

Keas.Mvc/ClientApp/pages/assets/AssetNav.tsx
sprucely: Oh, now I remember. I was testing the hot reloading and forgot to undo it.

sprucely:
Latest commit bdb4c57 implements a IConfigureNamedOptions to serve as an interceptor for CAS config options. It's a good pattern to keep in mind, particularly when you need to inject dependencies before all dependencies have been configured.

Keas.Mvc/Controllers/Api/KeySerialsController.cs
```diff
@@ -298,7 +307,6 @@ public async Task<IActionResult> Revoke(int id)
             // clear out assignment
             var assignment = keySerial.KeySerialAssignment;
             _context.KeySerialAssignments.Remove(assignment);
-            keySerial.KeySerialAssignmentId = null;
```
jSylvestre, line 301: @sprucely This isn't need anymore because the remove null's it out?

Keas.Mvc/Controllers/PersonAdminController.cs
```diff
@@ -190,10 +191,12 @@ public async Task<IActionResult> UploadPeople(PeopleImportModel model)
             var userName = User.GetNameClaim();
             var team = await _context.Teams.FirstAsync(t => t.Slug == Team);
             using (var reader = new StreamReader(model.File.OpenReadStream()))
-            using (var csv = new CsvReader(reader))
+            using (var csv = new CsvReader(reader, new CsvConfiguration(CultureInfo.InvariantCulture)
```
jSylvestre, line 194: @sprucely Not a big deal, but I'd put the "new CsvCo..." on a new line and indent.
Took me a minute to see that the using didn't get closed out.

Keas.Mvc/Controllers/Api/KeySerialsController.cs
sprucely: correct

PR #936, opened 2020-08-05
Do a few

jSylvestre:


srkirkland:
not sure the point, but the inline data stuff is cool for other tests

jSylvestre:
> not sure the point, but the inline data stuff is cool for other tests

Was just testing more than a couple. I'm not going to merge this.

jSylvestre:
Did not merge.

PR #939, opened 2020-08-05
Use like to compare each term

jSylvestre:


Keas.Mvc/Controllers/Api/SpacesController.cs
```diff
@@ -37,11 +38,17 @@ public async Task<IActionResult> SearchSpaces(string q)
 
             var queryWords = q.Split(" ").Where(w => !string.IsNullOrWhiteSpace(w)).ToArray();
 
-            var space = await _context.Spaces
-                .Where(x => orgIds.Contains(x.OrgId)
-                            && ((!string.IsNullOrWhiteSpace(x.BldgName) && queryWords.Contains(x.BldgName))
-                                || (!string.IsNullOrWhiteSpace(x.RoomName) && queryWords.Contains(x.BldgName))
-                                || (!string.IsNullOrWhiteSpace(x.RoomNumber) && queryWords.Contains(x.BldgName))))
+            var theQuery = _context.Spaces.Where(a => orgIds.Contains(a.OrgId)).AsQueryable();
```
srkirkland, line 41: Isn't is already Queryable?  Does it actually require AsQueryable?

sprucely: Not sure how that got in there.

sprucely: Oh, I see. Not my branch.

PR #946, opened 2020-08-11
Added a landing page for confirm similar to selectTeam

### LINKED ISSUE #888: Create Accept/Confirm landing page
Create a new landing page for confirming assignments without being based on a team.  Similar to the main home landing page, it'll pull up pending acceptance stuff and if you only have items in one team pending acceptance it'll redirect you straight there `/team/Confirm/Confirm`, otherwise it'll show you the teams in which you have pending stuff and make you choose which to go to.

This new page would then be the landing page for notification emails and possibly a future revision of the team home approvals header

wongband:
Copied the structure of SelectTeam rather than HomeViewModel since it was more needed for what the Issue wanted
closes #888


Keas.Mvc/Controllers/ConfirmController.cs
```diff
@@ -45,6 +45,32 @@ public async Task<IActionResult> MyStuff()
             return View(viewmodel);
         }
 
+        public async Task<IActionResult> Landing()
+        {
+            var team = await _securityService.GetPerson(Team);
```
srkirkland, line 50: @wongband do a check to see if `Team` is null/empty since there is no need to query the security service if `Team` is null.

wongband: Yea I can do that

wongband: If `Team` is null, where should I redirect the user?

srkirkland: `_securityService.GetPerson(Team);` will always return a null team if the passed in `Team` is null, so there is no reason to call it, just fall through to show the user a list of their teams,

PR #948, opened 2020-08-11
dev email sending

srkirkland:
update our email job to call a different method in dev/DEBUG and production.  This will allow us to send out sample emails (right now we have to query the db and hope some users match each email type).

I created the first sample email send method for reference


srkirkland:
@wongband looks good to me.  haven't tested but the code is what I was looking for.  This will set us up in a good position to update the emails for the app, as well as to add future email types

wongband:
@srkirkland is it good to merge?

PR #954, opened 2020-08-12
Jcs/api list inactive equipment

jSylvestre:
image


jSylvestre:
If you guys like this, I still have to update the tests

PR #955, opened 2020-08-13
Adding API endpoints

jSylvestre:


jSylvestre:
Might want to wait for Bryan's PR as it will probably have conflicts

Keas.Mvc/Controllers/Api/AccessController.cs
```diff
@@ -75,19 +75,43 @@ public async Task<IActionResult> List()
...
             return Json(accessList);
         }
+
+        /// <summary>
+        /// Lists all Access items that have been deleted for the team
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet]
+        [ProducesResponseType(typeof(IEnumerable<Access>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> ListInactive()
```
srkirkland, line 85: @jSylvestre I think you are probably better off here doing an overload like you did in the report controller.  Just add an param to our existing `List()` method that defaults `includeInactive == false`.

Keas.Mvc/Controllers/Api/AccessController.cs
```diff
@@ -213,5 +237,43 @@ public async Task<IActionResult> Delete(int id)
...
+        }
+
+        /// <summary>
+        /// Return all history records
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        [HttpGet("{id}")]
+        [ProducesResponseType(typeof(IEnumerable<History>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetFullHistory(int id)
```
srkirkland, line 268: do we inactivate history?  And if so, do we really want to return inactive history?  I don't see why this is needed

Keas.Mvc/Controllers/Api/AccessController.cs
jSylvestre: @srkirkland I'm ok with that, but I was thinking it would be useful to be able to see just inactive stuff.

Keas.Mvc/Controllers/Api/AccessController.cs
jSylvestre: @srkirkland Not the history, but the access that has the history.

PR #960, opened 2020-08-13
Added KeyInfo model for the SwaggerResponses

### LINKED ISSUE #957: Keys swagger documentation
Key List returns a object that isn't a key.
It will need to have a model created that matches what it returns, then use that instead of below.
[ProducesResponseType(typeof(IEnumerable<Key>), StatusCodes.Status200OK)]

There are probably a couple other of these.

wongband:
closes #957


Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -160,7 +158,7 @@ public async Task<IActionResult> Create([FromBody]CreateKeyViewModel model)
         }
 
         [HttpPost("{id}")]
-        [ProducesResponseType(typeof(Key), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(KeyInfo), StatusCodes.Status200OK)]
```
jSylvestre, line 161: I didn't test it, but this looks like it is returning a key, not keyInfo

Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -192,7 +190,7 @@ public async Task<IActionResult> Update(int id, [FromBody]UpdateKeyViewModel mod
...
         public async Task<IActionResult> Delete(int id)
```
jSylvestre, line 194: returns null?

Keas.Core/Domain/KeyData.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace Keas.Core.Domain
+{
+    public class KeyInfo
+    {
+        [Required]
+        public int id { get; set; }
+        
+        public Key Key { get; set; }
```
jSylvestre, line 12: List isn't returning a key. It is returning part of a key

Keas.Mvc/Controllers/Api/KeysController.cs
wongband: Thinking about it now since this endpoint in particular returns `null` I think we should not have a ResponseType for it?

Keas.Core/Domain/KeyData.cs
wongband: This should be fixed when I move it to the models folder

Keas.Mvc/Controllers/Api/KeysController.cs
wongband: Should be fixed when I push the latest commit

Keas.Mvc/Controllers/Api/KeysController.cs
jSylvestre: @sprucely What do you think? ☝️

sprucely: Oops, didn't see this last week. There's an overload of ProducesResponseType that takes just the status code. Should probably at least use that.

Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -60,7 +58,7 @@ select new
 
         // List all keys for a team
         [HttpGet]
-        [ProducesResponseType(typeof(IEnumerable<Key>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(IEnumerable<KeyInfo>), StatusCodes.Status200OK)]
         public async Task<IActionResult> List()
```
jSylvestre, line 62: @wongband This does return a key, but we actually cut out a bunch of stuff.

I'm thinking that this should have a new model, maybe call KeyPartial?
                key = new Key
                {
                    Id = r.Id,
                    Code = r.Code,
                    Name = r.Name,
                    Notes = r.Notes,
                    TeamId = r.TeamId,
                    Tags = r.Tags
                },
                id = r.Id,
                SpacesCount = r.SpacesCount,
                SerialsInUseCount = r.SerialsInUseCount,
                SerialsTotalCount = r.SerialsTotalCount

wongband: Yea that works as well

Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -91,7 +89,7 @@ public async Task<IActionResult> List()
 
         // list all keys for a space
         [HttpGet]
-        [ProducesResponseType(typeof(IEnumerable<Key>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(IEnumerable<KeyInfo>), StatusCodes.Status200OK)]
         public async Task<IActionResult> GetKeysInSpace(int spaceId)
```
jSylvestre, line 93: @wongband This returns:

{
    "key": {
      "code": "AEXMPLE2",
      "keyXSpaces": [
        {
          "id": 23,
          "keyId": 3562,
          "space": null,
          "spaceId": 3637
        }
      ],
      "serials": [
        {
          "number": "1",
          "status": "Destroyed",
          "keyId": 3562,
          "keySerialAssignment": null,
          "keySerialAssignmentId": null,
          "title": "1",
          "id": 8595,
          "name": "1",
          "notes": "asd",
          "tags": "",
          "team": null,
          "teamId": 21,
          "active": true
        },
        {
          "number": "Test2",
          "status": "Active",
          "keyId": 3562,
          "keySerialAssignment": null,
          "keySerialAssignmentId": null,
          "title": "Test2",
          "id": 8596,
          "name": "Test2",
          "notes": "",
          "tags": "",
          "team": null,
          "teamId": 21,
          "active": true
        }
      ],
      "title": "Destroyed key. Delete this line.",
      "id": 3562,
      "name": "Destroyed key. Delete this line.",
      "notes": "asdf",
      "tags": "Imported",
      "team": null,
      "teamId": 21,
      "active": true
    },
    "id": 3562,
    "serialsTotalCount": 1,
    "serialsInUseCount": 0,
    "spacesCount": 0
  }

Keas.Mvc/Controllers/Api/KeysController.cs
```diff
@@ -123,7 +121,7 @@ select new
         }
 
         [HttpPost]
-        [ProducesResponseType(typeof(Key), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(KeyInfo), StatusCodes.Status200OK)]
         public async Task<IActionResult> Create([FromBody]CreateKeyViewModel model)
```
jSylvestre, line 125: @wongband This just returns 

            var key = new Key()
            {
                Name = model.Name.Trim(),
                Code = model.Code.Trim(),
                Notes = model.Notes,
                Tags = model.Tags,
            };
Plus the team.
I'm not sure it is worthwhile to create a partial for this... Maybe just switch back to Key?

jSylvestre: @wongband I think this is the way to go, but Create (Probably the other http posts too) doesn't return a counts that you have in the KeyPartial. But lets wait for @srkirkland . I'd like to get his input too.

wongband: @jSylvestre sounds good.

Keas.Core/Models/KeyPartial.cs
```diff
@@ -0,0 +1,27 @@
...
+        public int Id { get; set; }
+        
+        public KeyDTO Key { get; set; }
+
+        public int SpacesCount { get; set; }
+
+        public int SerialsInUseCount { get; set; }
+ 
+        public int SerialsTotalCount { get; set; }
+    }
```
sprucely, line 26: Is there a specific reason for splitting this info into two classes? If not, I'd recommend merging them into a single flat class.

wongband: I'm using a DTO for the Key since there are some elements of the Key that we don't need to return back.

sprucely: But why not have your KeyDTO include SpacesCount, SerialsInUseCount and SerialsTotalsCount?

wongband: @jSylvestre wanted the model to look like:
```
key = new Key
{
Id = r.Id,
Code = r.Code,
Name = r.Name,
Notes = r.Notes,
TeamId = r.TeamId,
Tags = r.Tags
},
id = r.Id,
SpacesCount = r.SpacesCount,
SerialsInUseCount = r.SerialsInUseCount,
SerialsTotalCount = r.SerialsTotalCount
```
So I left them outside of the DTO. Is it better to have them inside @sprucely ?

Keas.Mvc/Controllers/Api/KeysController.cs
wongband: @srkirkland any thoughts about this?

srkirkland: Agree with @jSylvestre here, looking at the body of the method it appears to just be returning the `new Key()` object so we should be able to just use the response type of `Key` as it was

Keas.Core/Models/KeyPartial.cs
sprucely: I think it's cleaner, but @jSylvestre may have a reason for keeping them separate.

Keas.Mvc/Controllers/Api/KeysController.cs
jSylvestre: What I don't like, but don't have a good solution for, is that the key documentation in swagger also has all the other "includes" which are not set in this case.

sprucely: If we don't want to go the DTO route just for the sake of swagger documentation, maybe we can figure out a custom attribute solution to tell SwaggerGen to filter out unwanted properties.

wongband: @sprucely What do you think about using JsonIgnore to not show certain variables in the Swagger UI?

Keas.Core/Models/KeyPartial.cs
wongband: @jSylvestre Thoughts about this?

jSylvestre: @wongband I don't have an opinion on it. Maybe check with @srkirkland

srkirkland: I don't see why you would need the KeyDTO part since Key already covers the same properties, but you do need a new object with Key as a prop in order to add on the additional fields

wongband: @srkirkland The reason why we added the KeyDTO is because the `Key` Class has attributes that are not needed such as KeyXSpaces

srkirkland: Ok then I'm for it -- maybe put the models into a folder "Api" or some such so we know it's only used for that purpose.

wongband: Okay sounds good. Should we add: 
```
        public int SpacesCount { get; set; }

        public int SerialsInUseCount { get; set; }

        public int SerialsTotalCount { get; set; }
```
in the KeyDTO?

srkirkland: From what I can tell, that won't produce the correct output structure, but if it does then I'm for it.  My guess is you are going to need a class KeyWithCounts class which has the KeyDTO as one property.

wongband: Since the return data looks like:
```
{
  "key": {
    "code": "ER",
    "keyXSpaces": null,
    "serials": null,
    "title": "Test",
    "id": 23,
    "name": "Test",
    "notes": null,
    "tags": "Student",
    "team": null,
    "teamId": 10,
    "active": true
  },
  "id": 23,
  "spacesCount": 0,
  "serialsInUseCount": 2,
  "serialsTotalCount": 2
}
```
I think it looks right to have them outside

Keas.Mvc/ClientApp/components/Keys/KeyContainer.tsx
```diff
@@ -386,6 +386,7 @@ export default class KeyContainer extends React.Component<IProps, IState> {
         body: JSON.stringify(request),
         method: 'POST'
       });
+      console.log(key);
```
srkirkland, line 389: dump this

wongband: whoops accidentally left that

srkirkland:
> So `KeyDTO `and `KeyResponse `are only referenced in the attributes and not actually returned?
> 
> I feel like we've gone in a couple circles on this. Should we just ditch these types and place `Key` in the `[ProducesResponseType]` attributes and later worry about making the actual returned data consistent?

I would vote for that

wongband:
Im confused with what you mean for returned?

PR #963, opened 2020-08-14
New table for team

jSylvestre:


Keas.Mvc/Handlers/ApiKeyMiddleware.cs
```diff
@@ -45,16 +45,18 @@ public async Task Invoke(HttpContext context, ApplicationDbContext dbContext, IO
...
+            var teamApi = dbContext.TeamApiCodes.Include(a => a.Team).FirstOrDefault(a => a.ApiCode == headerGuidValue);
+            if (teamApi == null || teamApi.Team == null)
             {
                 // no team found with your auth token, fail
                 context.Response.StatusCode = 401; //UnAuthorized
                 await context.Response.WriteAsync(InvalidApiKeyMessage);
                 return;
             }
 
+            var teamAccess = teamApi.Team;
```
srkirkland, line 57: I think you could do this whole things with just replacing the first line with `var teamAccess = dbContext.Teams.FirstOrDefault(t => t. TeamApiCode.ApiCode == headerGuidValue);`

Keas.Sql/dbo/Tables/TeamApiCodes.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE [dbo].[TeamApiCodes]
+(
+	[Id]     INT            IDENTITY (1, 1) NOT NULL,
+	[TeamId] INT            NOT NULL,
+	[ApiCode] UNIQUEIDENTIFIER NOT NULL,
+	CONSTRAINT [FK_TeamApiCode_Teams_TeamId] FOREIGN KEY ([TeamId]) REFERENCES [dbo].[Teams] ([Id]) ON DELETE CASCADE,
```
jSylvestre, line 6: @srkirkland Does this look right?

Keas.Mvc/Handlers/ApiKeyMiddleware.cs
```diff
@@ -45,8 +45,7 @@ public async Task Invoke(HttpContext context, ApplicationDbContext dbContext, IO
             }
 
             // lookup team that this API key has acess to
-            var teamAccess = dbContext.Teams.FirstOrDefault(t => t.ApiCode == headerGuidValue);
-
+            var teamAccess = dbContext.Teams.Include(a => a.TeamApiCode).FirstOrDefault(a => a.TeamApiCode != null && a.TeamApiCode.ApiCode == headerGuidValue);
```
srkirkland, line 48: You don't need the .Include because you aren't using it in the return. Also I don't think you need the null check since it's ignored in the generated sql but I don't think it hurts anything.

Keas.Sql/dbo/Tables/TeamApiCodes.sql
srkirkland: Looks good but it's probably worth putting an index on api code since that's what we are searching on.

jSylvestre:
Test and Prod databases have been updated

PR #966, opened 2020-08-17
Person with their assignments

jSylvestre:


Keas.Core/Domain/EquipmentAssignment.cs
```diff
@@ -1,6 +1,6 @@
 namespace Keas.Core.Domain
 {
     public class EquipmentAssignment : AssignmentBase {
-
+        public Equipment Equipment { get; set; }
```
srkirkland, line 4: @jSylvestre if you are going to add these you should also add the related ID properties, even if you aren't going to use them.  So like add `public int EquipmentId {..}`.

PR #967, opened 2020-08-18
Fixes

jSylvestre:
remove unused var


Keas.Mvc/ClientApp/components/Shared/ReactTable.tsx
```diff
@@ -7,10 +7,15 @@ import {
   usePagination
 } from 'react-table';
 import { ColumnFilterHeaders, DefaultColumnFilter } from './Filtering';
-import { Pagination, PaginationItem, PaginationLink } from 'reactstrap';
```
jSylvestre, line 10: This is the only non test code change. Everything else is prettier formatting

PR #971, opened 2020-08-20
Added history for assigning keys to spaces

### LINKED ISSUE #959: Key History
I'm not seeing any history for a key when a space is associated with it.

image

wongband:
closes #959

### DIFF
--- a/Keas.Mvc/Controllers/Api/KeysController.cs
+++ b/Keas.Mvc/Controllers/Api/KeysController.cs
@@ -281,6 +281,7 @@ namespace Keas.Mvc.Controllers.Api
             _context.KeyXSpaces.Add(association);
             //await _eventService.TrackAssignKeySerial(serial);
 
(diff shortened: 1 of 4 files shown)


wongband:
After keeping track of history for assigning keys to spaces. Should we also keep track of when keys get dissociated from spaces @srkirkland @jSylvestre?

srkirkland:
@wongband Yea I agree adding dissociate would be good as well. I'm not sure about the notification service though if we want to do notifications for this

wongband:
Should I remove the notification? I only added it because the other event services did it as well, so I wanted to follow the consistently of that.

jSylvestre:
> @wongband Yea I agree adding dissociate would be good as well. I'm not sure about the notification service though if we want to do notifications for this

My vote would be not to notify

wongband:
Okay so no notifications for both associate and disassoicate right?

PR #978, opened 2020-08-24
Got rid of loosing filtering for equipment table

### LINKED ISSUE #962: Loosing filtering
Steven Barkey 7:52 AM

After filtering equipment by last name, clicking on the details button for a piece of equipment clears the filtering.  (likely happens with all column filters).

wongband:
closes #962 

I removed the `fetchDetails` function since the selected equipment passed in the component is the exact same as the equipment returned from `fetchDetails`. I think that this is not necessary and it does not remove the filter as well.

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentDetails.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentDetails.tsx
@@ -24,7 +24,6 @@ export default class EquipmentDetails extends React.Component<IProps, {}> {
     if (!this.props.selectedEquipment) {
       return;
     }
-    this._fetchDetails(this.props.selectedEquipment.id);
   }
 
   public render() {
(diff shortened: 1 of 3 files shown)


sprucely:
> 
> 
> You have removed the error checking.
> I don't really like that. But I'll let Scott decide.

@jSylvestre Some of the error checking in that removed method can be found at https://github.com/ucdavis/Peaks/blob/79c6869edaca67a7c307830f54116248465919aa/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx#L79-L87

But, yeah, it probably would be good to add in the toast with the "Not found" message.

jSylvestre:
> @jSylvestre Some of the error checking in that removed method can be found at

@sprucely That is getting the list of equipment. Not the individual equipment I believe. Two different actions.

wongband:
Should we do error checking if the `selectedEquipment` is null? From the original code it always fetches for the equipment regardless if the `selectedEquipment` is not null.

sprucely:
> 
> 
> > @jSylvestre Some of the error checking in that removed method can be found at
> 
> @sprucely That is getting the list of equipment. Not the individual equipment I believe. Two different actions.

Yes, it's getting the list of equipment, but it's an element of that list which is being passed to EquipmentDetails.

PR #979, opened 2020-08-26
include brandId in envelope if available

srkirkland:


Keas.Mvc/Services/DocumentSigningService.cs
```diff
@@ -57,6 +57,10 @@ public async Task<EnvelopeSummary> SendTemplate(string signerEmail, string signe
 
             envelope.Status = "sent";
 
+            if (!string.IsNullOrEmpty(_documentSigningSettings.BrandId)) {
+                envelope.BrandId = _documentSigningSettings.BrandId;
+            }
```
sprucely, line 62: Is there a reason not to use IsNullOrWhitespace?

srkirkland: figured i'd save 0.0000000001 nano seconds per call since we control the appsettings and i know it'll either be there or empty :0

PR #980, opened 2020-08-26
Fixed table not updating when revoking and assigning

wongband:
Fixed bug when table would not update when assigning or revoking equipment and keyserials


srkirkland:
@jSylvestre yea i figured it would but at least the rest isn't broken -- i think loosing the filter is the better outcome at the moment

PR #981, opened 2020-08-26
Updated ReactTable so that columns are never autoresetted

wongband:


jSylvestre:
@wongband This is just the sort, not the filter?

wongband:
@srkirkland realized that when someone was typing something in the input search bar the filters would reset. So by adding the `autoResetSortBy:false` we make it so that the columns dont auto rest when applying input filters

PR #982, opened 2020-08-27
Jcs/document report

jSylvestre:


Keas.Mvc/Keas.Mvc.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.5" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.1.5" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="3.1.5" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="3.1.4" />
```
sprucely, line 26: This one snuck in again

srkirkland: yea seems VS adds that automatically when you do certain scaffold-able options, like adding a view.  I double checked and it restores fine on mac and doesn't cause any issues so we might as well just keep it since it'll try to come back itself anyway

Keas.Mvc/Services/ReportService.cs
```diff
@@ -291,5 +293,26 @@ public async Task<IList<KeyReportModel>> Keys(Team team, string teamSlug, bool i
...
+            }
+
+            var people = await _context.People.AsNoTracking().Include(a => a.Documents)
+                .Where(a => a.TeamId == team.Id && a.Documents.Any(x => x.Active && x.Status != "Completed"))
+                .Select(s => new IncompleteDocumentReportModel()
+                {
+                    Name = s.Name,
+                    Email = s.Email,
+                    PersonId = s.Id,
+                    IncompleteDocumentCount = s.Documents.Count(w => w.Active && w.Status != "Completed"),
```
srkirkland, line 311: @jSylvestre how about we change to status == "sent".  Since there are other completions like Cancelled, Voided, Denied, etc that are possible end states

jSylvestre: @srkirkland So, if it is anything other than sent or completed, will someone go in and set it to inactive? (Delete it).
My thought is the admin want to know about all that are not completed. And if there is some other status, they should "fix" it.

srkirkland: Those statuses all come from docusign.  An example would be if the document isn't signed within 120 days it'll go to some Void status or similar.  There is nothing that can be done to fix it, since that is the correct final state.  I also think there might be some way for the signer to reject signing the document, which is a similar case.

jSylvestre:
image

Keas.Mvc/Services/ReportService.cs
jSylvestre: @srkirkland  If a "delete document" action is added to the documents container, then I think this will work well.
We would want to report all people who have not signed a document. This would include the Void and rejected status.
In these cases, the document admin should act on these cases. Probably by deleting the document, then adding it again. Unless it was done in error.

srkirkland: Yea I think that sounds good, though I'm in no rush to add the document delete functionality to my nice clean table ;0

PR #989, opened 2020-08-28
Document Delete API endpoint

jSylvestre:


Keas.Mvc/Services/HistoryService.cs
```diff
@@ -657,6 +658,22 @@ public async Task<History> DocumentStatusChange(Document document)
             _context.Histories.Add(historyEntry);
             return historyEntry;
         }
+
+        public async Task<History> DocumentDeleted(Document document)
+        {
+            var person = await _securityService.GetPerson(document.TeamId);
```
sprucely, line 664: Shouldn't document deletion history store the id of the person performing the deletion and not the generic Team person?

Oh, wait, I might be misunderstanding how securityService.GetPerson works.

Yeah, it looks the choice of real vs generic team user depends on authentication method. So deletions by token-authenticated calls will only show a team user.

Keas.Mvc/Controllers/Api/DocumentController.cs
```diff
@@ -119,5 +119,38 @@ public async Task<IActionResult> Create([FromBody] Document document)
...
+
+            if (!document.Active)
+            {
+                return BadRequest(ModelState);
+            }
+
+            using (var transaction = _context.Database.BeginTransaction())
+            {
+
+                _context.Documents.Update(document);
```
srkirkland, line 144: Do you need this line?  Shouldn't it be implicit because you grabbed the object already and didn't do noTracking so it should be tracked already.

jSylvestre: I copied from equipment.

PR #990, opened 2020-08-28
send reminders for unsigned documents 

srkirkland:


Keas.Mvc/Services/DocumentSigningService.cs
```diff
@@ -22,6 +22,7 @@ public interface IDocumentSigningService
 
     public class DocumentSigningService : IDocumentSigningService
     {
+        private static readonly string ReminderFrequency = "7"; // send reminders once a week
```
sprucely, line 25: Should there be a way to make this configurable per template and/or document? I can envision some documents being more urgent than others.

srkirkland: Jason and I did talk about that specifically and we decided to keep it standard to reduce work+confusion.  But I could definitely see that being something we add later.

PR #993, opened 2020-08-31
Rest of Active menu Nav items

jSylvestre:


jSylvestre:
> Verified it's applying the nav-active style appropriately (Though not sure what the purpose of the style is. I don't see any visible difference for the active tab)

Cal is going to do a style for us that indicates what is currently selected on the menu.

PR #994, opened 2020-08-31
Updated UI to edit Access Assignments

### LINKED ISSUE #938: Access Update Assingment
Access is a little different from the other, but if you want to update the expiration date, you have to revoke the person, then add them back.
There is no "Update Assignment"

wongband:
closes #938

### DIFF
--- a/Keas.Mvc/Controllers/Api/AccessController.cs
+++ b/Keas.Mvc/Controllers/Api/AccessController.cs
@@ -152,23 +152,34 @@ namespace Keas.Mvc.Controllers.Api
                 return BadRequest(ModelState);
             }
 
-            var access = await _context.Access.Where(x => x.Team.Slug == Team)
-                .Include(x => x.Assignments).SingleAsync(x => x.Id == accessId);
+            var assignment = await _context.AccessAssignments.Where(x => x.AccessId == accessId && x.PersonId == personId)
+                .SingleAsync();
 
-            var accessAssignment = new AccessAssignment
+            if (assignment != null) 
             {
-                AccessId = accessId,
-                PersonId = personId,
-                RequestedById = User.Identity.Name,
-                RequestedByName = User.GetNameClaim(),
-                ExpiresAt = DateTime.Parse(date),
-            };
+                assignment.ExpiresAt = DateTime.Parse(date);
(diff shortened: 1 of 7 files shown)


jSylvestre:
This is different (maybe better but not as obvious) than how it is done when you assign new access.
Your change if you manually try to change the year:
image

The way it is done/looks when you assign it with a past date:
image

wongband:
Ohhh I see where I went wrong

jSylvestre:
@wongband Almost. Just need to disable the update when it is invalid:

image

wongband:
Sounds good.

Keas.Mvc/Controllers/Api/AccessController.cs
```diff
@@ -192,6 +192,22 @@ public async Task<IActionResult> Update([FromBody] Access access)
 
         }
 
+        [HttpPost("{id}")]
+        [ProducesResponseType(typeof(AccessAssignment), StatusCodes.Status200OK)]
+        public async Task<IActionResult> UpdateAssignment([FromBody] AccessAssignment accessAssignment)
```
srkirkland, line 197: @wongband see if you can use the existing `Assign` method here, like we do in equipment or other places. See https://github.com/ucdavis/Peaks/blob/master/Keas.Mvc/Controllers/Api/EquipmentController.cs#L210.  You'll probably have to add an IF statement to see if it's an existing assignment or just an update.  Also, we'll want to track this action like we do when the assignment is first updated.  Again check out the equipment controller a good example.

srkirkland:
> This pull request **introduces 1 alert** when merging [bcb97a4](https://github.com/ucdavis/Peaks/commit/bcb97a404dffd1d355f6f2b8a8a2eb3c727be260) into [772440a](https://github.com/ucdavis/Peaks/commit/772440accf9543405fe5416bf26ec6a10331caca) - [view on LGTM.com](https://lgtm.com/projects/g/ucdavis/Peaks/rev/pr-14954f5c8faf8ced2a05c64d028097a676ae649b)
> 
> **new alerts:**
> 
> * 1 for Unused or undefined state property

Also fix this lgtm issue -- you have `date` in the state but you never use it

srkirkland (approved):
Code looks good and the feature seems to work well 👍

PR #1003, opened 2020-09-02
Limit swagger content type to application/json

### LINKED ISSUE #1002: Swagger / XML
Make it so swagger only shows request content types for application/json

If this is difficult, we could add support for XML in the API, but we would prefer not to do this.

sprucely:
closes #1002 

Made it for both request and response content types for all actions on all api controllers. Could be controlled on a more finely-grained basis if necessary.

### DIFF
--- a/Keas.Mvc/Controllers/Api/DocumentController.cs
+++ b/Keas.Mvc/Controllers/Api/DocumentController.cs
@@ -16,6 +16,8 @@ namespace Keas.Mvc.Controllers.Api
     [Authorize(Policy = AccessCodes.Codes.DocumentMasterAccess)]
     [ApiController]
     [Route("api/{teamName}/documents/[action]")]
+    [Consumes("application/json")]
+    [Produces("application/json")]
     public class DocumentsController : SuperController
     {
         private readonly ApplicationDbContext _context;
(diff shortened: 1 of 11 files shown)


Keas.Mvc/Controllers/Api/DocumentController.cs
```diff
@@ -16,6 +16,8 @@ namespace Keas.Mvc.Controllers.Api
     [Authorize(Policy = AccessCodes.Codes.DocumentMasterAccess)]
     [ApiController]
     [Route("api/{teamName}/documents/[action]")]
+    [Consumes("application/json")]
```
jSylvestre, line 19: @srkirkland Does the Document controller need XML?

jSylvestre (approved):
If the document one is OK with @srkirkland  then it looks good to me.

Keas.Mvc/Controllers/Api/DocumentController.cs
srkirkland: nope, just the WebhookController, which isn't part of our swagger API surface so this change should be fine

srkirkland (approved):

PR #1005, opened 2020-09-03
Jcs/doc api fix

jSylvestre:


Keas.Mvc/Controllers/Api/DocumentController.cs
```diff
@@ -34,6 +33,8 @@ public DocumentsController(ApplicationDbContext context, IDocumentSigningService
         // List all documents for this team
         [HttpGet]
         [ProducesResponseType(typeof(IEnumerable<Document>), StatusCodes.Status200OK)]
+        [Produces(MediaTypeNames.Application.Json)]
+        [Consumes(MediaTypeNames.Application.Json)]
         public async Task<IActionResult> List()
```
sprucely: For actions like this and ones that only take query or path parameters, should we not include a Consumes json attribute?

sprucely: Just saw Scott's comment on slack.

srkirkland (approved):

Keas.Mvc/Controllers/Api/DocumentController.cs
```diff
@@ -10,14 +10,14 @@
...
 
 namespace Keas.Mvc.Controllers.Api
 {
     [Authorize(Policy = AccessCodes.Codes.DocumentMasterAccess)]
     [ApiController]
     [Route("api/{teamName}/documents/[action]")]
-    [Consumes("application/json")]
-    [Produces("application/json")]
+    [Produces(MediaTypeNames.Application.Json)]
     public class DocumentsController : SuperController
```
sprucely, line 21: I'm not sure it's good to have Produces at the controller level for this controller. The Get action shows it produces both application/pdf and application/json. But I don't see how it can produce json.

jSylvestre: @sprucely It was Scott's suggestion.

PR #1006, opened 2020-09-03
clear build warnings about deadlocks

srkirkland:
switching to render partial async


sprucely (approved):
Didn't know about the `@await` keyword for views.

srkirkland:
@sprucely we are actually already using it in other parts of the app.  I was originally going to go with the newer tag helper `<partial name="abc" />` instead, which is async by default, but since we are using async partials elsewhere I just figured I'd standardize instead.

PR #1011, opened 2020-09-09
Moved modal to KeyDetailContainer to prevent going back to the KeyList

### LINKED ISSUE #1008: Edit Key Routing
To edit access, you need to click on details.
When you click "Edit Key", the edit dialog appears, but behind that the page turns to the Key list.

It should remain and update the details page

wongband:
closes #1008

### DIFF
--- a/Keas.Mvc/ClientApp/components/Keys/EditKey.tsx
+++ b/Keas.Mvc/ClientApp/components/Keys/EditKey.tsx
@@ -1,151 +1,123 @@
 import * as React from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { Button, Modal, ModalBody, ModalFooter } from 'reactstrap';
-import { Context } from '../../Context';
 import { IKey, keySchema } from '../../models/Keys';
 import { IValidationError, yupAssetValidation } from '../../models/Shared';
 import KeyEditValues from './KeyEditValues';
 
 interface IProps {
-  onEdit: (key: IKey) => void;
   modal: boolean;
-  closeModal: () => void;
   selectedKey: IKey;
   searchableTags: string[];
+  onEdit: (key: IKey) => void;
+  closeModal: () => void;
   checkIfKeyCodeIsValid: (code: string, id: number) => boolean;
 }
 
-interface IState {
-  error: IValidationError;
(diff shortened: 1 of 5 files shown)


Keas.Mvc/ClientApp/components/Keys/EditKey.tsx
```diff
@@ -1,151 +1,123 @@
...
-          ...prevState.key,
-          [property]: value
-        }
-      }),
-      this._validateState
-    );
+  const { searchableTags } = props;
+  const changeProperty = (property: string, value: string) => {
+    key[property] = value;
+    setKey(key);
```
srkirkland, line 47: @wongband this is not actually the same thing, and I'm not sure if it'll create error or not but you should verify your useEffect is being called if you go this way.  Ex: `key[property] = value; setKey(key);` like you did means that `key` is the same object reference the whole time.  Where as `setKey({...key, [property]: value})` creates a whole new object to put into state.  It used to be that you always wanted to replace your object when modifying state, or else the change detection in setState() wouldn't always work properly, especially for nested objects.... Anyway, mainly I mention this so you can be extra sure than when replacing a fully-new object with a simply modified object that it doesn't change behavior and that state is actually updated and related useEffect(()=>{}, [key]) and other hooks still work.

wongband: I just checked with:
```
key[property] = value;
setKey(key);
```
and
```
    setKey({...key, [property]: value})
```
Both methods calls `useEffect`. I'm assuming that we should go with the second way since that is the standard way to do it?

srkirkland: @wongband yes I think so too, especially when we get into nested objects I think it'll be a bigger deal.  I'll read more on the topic (of immutable w/ hooks) but I agree I think standardizing on the second way is a good idea

wongband: Okay sounds good. I'll change it to that right now

jSylvestre (approved):
Looks good to me from a testing point of view

Issue #1009, opened 2020-09-09
Group Level Key Report


srkirkland:
Need a report at the group level which shows who has keys to which spaces.   A lot like the existing "Keys With Spaces Report" except it should also show who has been assigned that key.  

Requested by Rob Kerner

PR #1012, opened 2020-09-10
Group report for Key Serials with Spaces

srkirkland:
There isn't a matching report in the team reports section yet.  This report only shows key serials which have assigned spaces, and it'll show who that key serial is assigned to, once for each space.


srkirkland:
closes #1009

jSylvestre (approved):
🚢

PR #1018, opened 2020-09-15
Converted more workstation sub components to use hooks

wongband:
Converted the following components to use hooks:
- WorkstationDetails
- EditWorkstation
- DeleteWorkstation
- SearchWorkstations
- WorkstationAssignmentValues

### DIFF
--- a/Keas.Mvc/ClientApp/components/Workstations/WorkstationDetails.tsx
+++ b/Keas.Mvc/ClientApp/components/Workstations/WorkstationDetails.tsx
@@ -1,4 +1,5 @@
 ﻿import * as React from 'react';
+import { useContext, useEffect } from 'react';
 import { toast } from 'react-toastify';
 import { Button, Modal, ModalBody } from 'reactstrap';
 import { Context } from '../../Context';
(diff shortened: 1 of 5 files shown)


Keas.Mvc/ClientApp/components/Workstations/WorkstationDetails.tsx
```diff
@@ -16,73 +17,73 @@ interface IProps {
...
-export default class WorkstationDetails extends React.Component<IProps, {}> {
-  public static contextType = Context;
-  public context!: React.ContextType<typeof Context>;
+const WorkstationDetails = (props: IProps) => {
+  const context = useContext(Context);
+  const workstation = props.selectedWorkstation;
 
-  public componentDidMount() {
-    if (!this.props.selectedWorkstation) {
+  useEffect(() => {
```
srkirkland, line 24: Are you sure you want this to run on every render?  I think maybe instead did you mean to add the empty dependency array `useEffect(()=>, [])` and have it run just once?  I feel like otherwise it's going to run over and over in a loop as the fetch method causes the parent props to change.  Maybe not a huge issue, but I think it's good practice to always provide the second argument to useEffect()

srkirkland (changes requested):

Keas.Mvc/ClientApp/components/Workstations/WorkstationDetails.tsx
wongband: Shoot forgot about that. I will change that right now.

jSylvestre (approved):

srkirkland (approved):

PR #1019, opened 2020-09-16
Converted sub components from People to use hooks

wongband:
Converted the following components to use hooks:
- AssignPerson
- BioContainer
- PeopleListItem
- PeopleTable
- PersonDetails
- SearchUsers


srkirkland (approved):
seems solid to me

jSylvestre (approved):

PR #1026, opened 2020-09-18
Converted some sub components in Access folder to use hooks

wongband:
Converted the following components:
- AccessList
- AccessListItem
- AccessTable
- AssignAccess
- RevokeAccess


sprucely (commented):
I only glanced at the code, but ran it locally and everything seems to work for me.

jSylvestre (changes requested):
When I create a new access, the list isn't updated unless I refresh the page.

wongband:
@jSylvestre Should be fixed in this push ^^^

jSylvestre (changes requested):
@wongband The list update appear to be fixed.
But.... When creating an access, the add tags (click on tags, then click on the tag) is mostly not working.

image

jSylvestre (approved):
@wongband Looks good to me, but wait for @srkirkland to take a look at the code.

srkirkland (approved):

PR #1027, opened 2020-09-21
Converted the rest of Access' sub components to use hooks

wongband:
Converted the following components:

- AccessEditValues 
- DeleteAccess
- SearchAccess

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useEffect } from 'react';
 import { Button, FormFeedback, FormGroup, Input, Label } from 'reactstrap';
 import { IAccess } from '../../models/Access';
 import { IValidationError } from '../../models/Shared';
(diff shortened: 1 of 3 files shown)


jSylvestre (approved):

Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
```diff
@@ -7,100 +8,97 @@ import SearchTags from '../Tags/SearchTags';
...
   onAccessUpdate?: (access: IAccess) => void;
 }
 
-export default class AccessEditValues extends React.Component<
-  React.PropsWithChildren<IProps>
-> {
-  constructor(props: React.PropsWithChildren<IProps>) {
-    super(props);
+const AccessEditValues = (props: IProps) => {
+  useEffect(() => {
```
srkirkland, line 19: What's the point of useEffect here?  If this is just running during every render you don't need the useEffect, and if it's just supposed to run once you need to add the second param

srkirkland (changes requested):

Keas.Mvc/ClientApp/components/Access/AccessEditValues.tsx
wongband: Oh shootI thought it was used in a `componentDidMount` when it was really in a constructor. I will remove the useEffect

srkirkland (approved):

PR #1038, opened 2020-10-02
Converted KeysContainer to use hooks

wongband:
Updated the following containers:
- KeyContainer
- KeySerialContainer
- AccessContainer
- AccessAssignmentContainer
- WorkstationContainer

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
@@ -76,7 +76,7 @@ const AssignmentContainer = (props: IProps) => {
 
     getAssignments();
   }, [
-    assignments.length,
+    assignments,
     context,
     params.id,
     props.person
--- a/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { RouteChildrenProps, withRouter } from 'react-router';
+import { useContext, useEffect, useState } from 'react';
+import {
+  RouteChildrenProps,
+  withRouter,
+  useHistory,
+  useParams
+} from 'react-router';
 import { toast } from 'react-toastify';
 import { Context } from '../../Context';
 import { IAccess, IAccessAssignment } from '../../models/Access';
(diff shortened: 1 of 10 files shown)


jSylvestre (approved):

Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
```diff
@@ -15,199 +16,112 @@ import { PermissionsUtil } from '../../util/permissions';
...
+        setAssignments(assignmentsData);
+        setSelectedAssignment(
+          assignmentsData.find(el => el.id === parseInt(params.id, 10))
+        );
+      }
+    };
+
+    getAssignments();
+  }, [
+    assignments.length,
```
srkirkland, line 79: why only if the length changes but not the assignments themselves (or at least the reference to?)

Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
```diff
@@ -273,27 +183,84 @@ class AssignmentContainer extends React.Component<IProps, IState> {
...
+  const isEditModalShown =
+    assetType === 'accessAssignment' && action === 'update';
+  const isAssignModalShown = assetType === 'access' && action === 'assign';
+
+  return (
+    <div>
+      {isRevokeModalShown && (
+        <RevokeAccess
+          assignment={selectedAssignment}
+          revoke={callRevoke}
```
srkirkland, line 217: Maybe as a future optimization but when you passalong a callback, like callRevoke or any of those similar methods, you should use the `useCallback` hook to prevent extra unnecessary network calls.  It might be worth running this with the network tab open to make sure it's not doing anything to weird with this code already.

srkirkland (approved):

Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
wongband: Before, eslint just complains that I need to include assignments.length, but I just changed it to assignments and it seems fine with that.

Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
wongband: I think we can take a look at this in a future branch since we pass around multiple callbacks into components.

PR #1045, opened 2020-10-21
Fixed Access edit bug

### LINKED ISSUE #1036: Edit Access after Assigning user
Select Access tab
Click on details
Assign new user
Edit access and save,
Generates error

wongband:
closes #1036

### DIFF
--- a/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Access/AccessAssignmentContainer.tsx
@@ -75,12 +75,7 @@ const AssignmentContainer = (props: IProps) => {
     };
 
     getAssignments();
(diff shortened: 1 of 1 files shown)


srkirkland (approved):

jSylvestre:
@srkirkland This is better than the error, but it doesn't update the person (when adding from the person tab) unless you refresh:

If you are OK with this quirk, then I'm ok with the overall improvement.

image

srkirkland:
well I'm ok with it because I don't think it's worth spending too much time to fix, and plus this PR is important to get in.  but we should document the issue if we haven't already and maybe @wongband can try to spend a bit of time figuring out the issue.  But I'd say merge this pr first

jSylvestre (approved):

PR #1054, opened 2020-11-04
Mess up a test to show build failure

srkirkland:
This is just for the demo I'm doing Nov 10th, i'll close without merging after that.


srkirkland:
presentation over, no need for this anymore

PR #1055, opened 2020-11-13
Refactor documents for multiple accounts

### LINKED ISSUE #1034: Docusign not available to other colleges
COE User:

image

sprucely:
Seems to be working for me. Was able to assign myself a document, sign it and see it get marked as completed in db.

The Team Docusign Account info on the Document Settings page is kind of ugly...
image

Didn't have it in me at the time to figure out a better layout. But it's functional. 🤷 

closes #1034


Keas.Mvc/Services/DocumentSigningService.cs
```diff
@@ -30,19 +33,33 @@ public class DocumentSigningService : IDocumentSigningService
 
         private readonly DocumentSigningSettings _documentSigningSettings;
 
+        private OAuth.UserInfo _userInfo;
+
         public DocumentSigningService(IOptions<DocumentSigningSettings> documentSigningSettings)
         {
-            _apiClient = _apiClient ?? new ApiClient();
+            _apiClient ??= new ApiClient();
```
srkirkland, line 40: eww

srkirkland (approved):
The code looks good, the docs will need to be updated and we'll probably want to put a link to that on the docusign create page

Keas.Mvc/Services/DocumentSigningService.cs
jSylvestre: Yeah, I hope that isn't the new and shiny. I just got used to ?? :)

sprucely: @srkirkland What docusign create page are you referring to?

srkirkland: The team admin one. It used to have some basic instructions but there's nothing on the new one.

jSylvestre: @sprucely I think @srkirkland means the ones here:
https://computing.caes.ucdavis.edu/documentation/peaks/docusign-documents

Because you updated the page in PEAKS I believe.

jSylvestre (changes requested):
Use capitalization?

image

jSylvestre (approved):

Keas.Mvc/Controllers/Api/DocumentController.cs
```diff
@@ -103,17 +114,19 @@ public async Task<IActionResult> Create([FromBody] Document document)
...
                 PersonId = document.PersonId,
                 TeamId = document.TeamId,
                 EnvelopeId = envelope.EnvelopeId,
-                TemplateId = document.TemplateId,
+                TemplateId = document.TemplateId,
+                AccountId = document.AccountId,
+                ApiBasePath = document.ApiBasePath,
                 Active = true,
-                Status = "sent",
+                Status = "Sent",
```
srkirkland, line 129: Does this match what docusign sends back via their API?  I'd like it to match whatever capitalization that is.

sprucely: I believe it does. But I'll double-check...

sprucely: It looks like the `Completed` status comes back capitalized, so assume it would be the same for `Sent`.

PR #1058, opened 2020-11-20
Show templates if any

jSylvestre:
image


srkirkland (approved):
might be worth detecting the state if they have no templates, and for empty catch blocks I like to not add in the `Exception e` part because that'll just result in a compiler warning for unused var

PR #1063, opened 2020-12-07
add logging to elastic search platform

srkirkland:
all existing logging stays, this is just adding one more sink to the new dev elasticsearch platform


sprucely (approved):
Looks okay to me. Where can I find the ElasticUrl setting?

srkirkland:
> Looks okay to me. Where can I find the ElasticUrl setting?

So unfortunately at the moment our caes firewall blocks access to elasticsearch since it's on a non http standard port.  So we can't use peaks locally along with ES.  But that said, I'll add the ES setting to peaks test and then let you know once I get the right keys figured out.

PR #1064, opened 2020-12-07
Update serilog with newer init method and also better log only pipeline requests

srkirkland:
By default aspnet routing is very verbose.  Switch to using the serilog request logging stuff, see: https://github.com/serilog/serilog-aspnetcore#request-logging.


sprucely (approved):
Nice. I would have been surprised if serilog didn't have some sort of method of intercepting log events.

PR #1066, opened 2020-12-09
start logging username and client ip info

srkirkland:


sprucely (approved):
I can see why you like Serilog. It seems to include everything but the kitchen sink.

PR #1073, opened 2021-01-14
Don't require content-type when post has no body

sprucely:


srkirkland (approved):

Keas.Mvc/Startup.cs
```diff
@@ -263,6 +262,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerF
             app.UseRouting();
             app.UseSession();
             app.UseAuthentication();
+            app.UseMiddleware<ApiKeyMiddleware>();
```
jSylvestre, line 265: @sprucely For this to work correctly it had to happen after after useAuthentication?
If yes, it might be good to add that as a comment

jSylvestre (approved):
Remind me when it is merged and I'll test these on test.

PR #1076, opened 2021-01-21
Changed bigfix variable names/files to managedSystem

wongband:


srkirkland (approved):
Everything looks good except the TeamAdminController -- you have to be careful what you change because as you can tell it's impossible to determine what has been changed in that file.

wongband:
I removed the TestOs for bigfix, is that okay?

srkirkland:
> I removed the TestOs for bigfix, is that okay?

yup

PR #1081, opened 2021-01-26
API Validation and cleanup

jSylvestre:


Keas.Mvc/Controllers/Api/WorkstationsController.cs
```diff
@@ -256,6 +256,12 @@ public async Task<IActionResult> Create([FromBody] Workstation workstation)
             if (!await _securityService.IsSpaceInTeam(Team, workstation.Space.Id))
             {
                 return BadRequest("Space not in Team");
+            }
+
+            //Validate User (including API user, has permissions)
+            if (!await _securityService.IsUserAllowed(Team, Role.Codes.SpaceMaster))
```
srkirkland, line 262: Why?  Doesn't the `[Authorize(Policy = AccessCodes.Codes.SpaceMasterAccess)]` attr on the controller already do this?  if not, shouldn't it be done there?

srkirkland (changes requested):

Keas.Mvc/Controllers/Api/WorkstationsController.cs
jSylvestre: @srkirkland Yes, you are correct.

srkirkland (approved):

PR #1094, opened 2021-02-24
Adjusted Computer Search by multiple fields

wongband:


Keas.Mvc/Services/ServiceNowService.cs
```diff
@@ -27,9 +27,15 @@ public ServiceNowService(IOptions<ServiceNowSettings> serviceNowSettings)
 
         public async Task<ServiceNowPropertyWrapper> GetComputersByName(string name)
```
srkirkland, line 28: You should change this to be something like `GetComputersByProperty(string property)` since it isn't just name anymore

Keas.Mvc/Services/ServiceNowService.cs
```diff
@@ -27,9 +27,15 @@ public ServiceNowService(IOptions<ServiceNowSettings> serviceNowSettings)
...
+            // Using ServiceNow's API we can chain requests with ^OR
+            string nameQuery = "?sysparm_query=hardware_u_device_nameLIKE" + name;
+            string userNameQuery = "^ORuser_user_nameLIKE" + name;
+            string ipAddressQuery = "^ORhardware_u_ip_addressLIKE" + name;
+            string macAddressQuery = "^ORhardware_u_mac_addressLIKE" + name;
+            string serialNumberQuery = "^ORhardware_serial_numberLIKE" + name;
             string endUrl = "&sysparm_display_value=true&sysparm_exclude_reference_link=true&sysparm_limit=5";
-            string fullUrl = _serviceNowSettings.ApiBasePath + urlAddOns + name + endUrl;
+            string fullUrl = _serviceNowSettings.ApiBasePath + nameQuery + userNameQuery +
+                ipAddressQuery + macAddressQuery + serialNumberQuery + endUrl;
```
srkirkland, line 38: use a stringbuilder to build up this string.  Any time you are concatinating strings more than once or twice a string builder is better to save on extra memory allocations: https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=net-5.0.

So you'd do like `var url = new StringBuilder(ApiBasePath);` and then just append one at a time, `url.Append(<namequery here>)` and at the end use `url.ToString()` to get out the desired string.

srkirkland (changes requested):

jSylvestre (approved):

srkirkland (approved):

PR #1095, opened 2021-03-03
Errror messages wil now be what ever is in the BadRequest errorr

### LINKED ISSUE #1091: Better error messges?
If you edit an equipment, and there is some error from the API, that info is not returned.

For example, if spaces are removed from the team, but they are used for an equipment, if you edit the equipment, and don't change the "bad" space, you just get an error occurred.

Try and capture the text in the bad request.

wongband:
closes #1091 

- Throw the res.json() if there is an error.

### DIFF
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -424,7 +424,9 @@ const EquipmentContainer = (props: IProps): JSX.Element => {
       toast.success('Equipment updated successfully!');
     } catch (e) {
       const errorMessage =
-        e.message === '' ? 'Error editing equipment' : e.message;
+        e.message === ''
+          ? 'Error editing equipment'
+          : `Error editing equipment, ${e.message}`;
       toast.error(errorMessage);
       throw new Error(); // throw error so modal doesn't close
     }
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -423,7 +423,8 @@ const EquipmentContainer = (props: IProps): JSX.Element => {
       );
       toast.success('Equipment updated successfully!');
     } catch (e) {
-      const errorMessage = e.message === "" ? "Error editing equipment" : e.message;
+      const errorMessage =
+        e.message === '' ? 'Error editing equipment' : e.message;
       toast.error(errorMessage);
       throw new Error(); // throw error so modal doesn't close
     }
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -423,7 +423,7 @@ const EquipmentContainer = (props: IProps): JSX.Element => {
       );
       toast.success('Equipment updated successfully!');
     } catch (e) {
-      toast.error('Error editing equipment.');
+      toast.error(e);
       throw new Error(); // throw error so modal doesn't close
     }
 
--- a/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
+++ b/Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
@@ -423,7 +423,8 @@ const EquipmentContainer = (props: IProps): JSX.Element => {
       );
       toast.success('Equipment updated successfully!');
     } catch (e) {
-      toast.error(e);
+      const errorMessage = e.message === "" ? "Error editing equipment" : e.message;
+      toast.error(errorMessage);
       throw new Error(); // throw error so modal doesn't close
     }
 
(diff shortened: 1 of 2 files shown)


jSylvestre (commented):
It is possible there is no text, so we may want to deal with that.

wongband:
That is true. What should we put if the text is empty?

jSylvestre (changes requested):
The toast doesn't appear at all if the BadRequest doesn't have any text supplied.
I'd suggest keeping the "Error editing equipment." if there is no returned text, and if there is change it to Error editing equipment: the error text

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
```diff
@@ -423,7 +423,7 @@ const EquipmentContainer = (props: IProps): JSX.Element => {
       );
       toast.success('Equipment updated successfully!');
     } catch (e) {
-      toast.error('Error editing equipment.');
```
jSylvestre, line 426: If you want to do it quick and easy, just do `toast.error('Error editing equipment. ' + e);`
But @srkirkland may have a better way.

jSylvestre: > If you want to do it quick and easy, just do `toast.error('Error editing equipment. ' + e);`
> But @srkirkland may have a better way.

Actually, I don't think that works if the BadRequest doesn't have text. I think it does this if you do that change:
image

Keas.Mvc/ClientApp/util/api.ts
```diff
@@ -16,7 +16,7 @@ const doFetch = async (
   });
 
   if (!res.ok) {
-    throw new Error(res.statusText);
+    throw await res.json();
```
srkirkland, line 19: I don't see any reason to believe that the error result will be JSON in all cases.  In fact I think in many cases it'll just be body text `res.text()`.  I think you should check the content type and wrap the `await res.xyz()` method in a try/catch with a fallback to sending the statusText or statusCode.

Keas.Mvc/ClientApp/components/Equipment/EquipmentContainer.tsx
srkirkland: yea you'd want to use `e.Message` instead of just `e` (since e includes stack trace and other stuff).

Keas.Mvc/ClientApp/util/api.ts
wongband: Do you have an example of what would returned in a BadRequest as json?

srkirkland: No and I wouldn't expect it to be JSON at all.  But you can make a sample endpoint that throws different errors and test them.  Make one that throws BadRequest with a message, one that throws a blank one, one that just divides by 0, etc.  Try out a few thrown exception cases and then you can see if the change to the `!res.ok` error condition of our fetch method that you are making will work in different cases.

jSylvestre:
Throw an exception:
image

return BadRequest("Fake message"); :
image

return BadRequest();
image

jSylvestre (changes requested):
See my comment with the 3 images
If an exception happens, it should just say Error editing equipment.
If a bad request happens with no text, good.
If a bad request happens with text, you should still include the "Error editing equipment" as well as the text from the bad request.

wongband:
@srkirkland The content type shows up as application/json when putting a string into a BadRequest, which is kinda misleading

jSylvestre (approved):
I'll let @srkirkland review the actual code, but the UI is working correctly.

srkirkland:
Transcript of relevant Slack message between Brian and myself.

>what about the case where !res.ok but the error code isn’t 500 or 400?  Seems like you should want to handle that case and throw an error
what if the result is a 404?  502?  Or just anything else?
After you fix/change that, I think it’s up to Jason and if he changes his review to an approve then you can merge since he’s leading this change
my thought would be there is no reason to check status, unless you want to do something like checking the status >400.  But I think the idea should be if !res.ok then try to grab the res.json() and if it’s a string, return it, otherwise just throw a generic error which includes the status code since that’s all the info you have.

jSylvestre:
There are 4 things I'm testing:
BadRequest()
BadRequest("Some Message")
BadRequest(ModelState)
Exception()

The code seemed to work as expected with those.

wongband:
@srkirkland I'll make some adjustments today to make it more general with other errors as well

wongband:
For this push I made it so that we have a try catch while unwrapping, `res.json()`. This way if we fail unwrapping it, we should throw an error. That should encapsulate an errors that throw exceptions like 500 errors. Otherwise if we successfully, unwrap  `res.json()` and its a string we will return that message back to the client.

Keas.Mvc/ClientApp/util/api.ts
```diff
@@ -1,4 +1,5 @@
 import 'isomorphic-fetch';
+import { errorMonitor } from 'stream';
```
srkirkland: @wongband remove this

srkirkland (approved):
looks good

PR #1102, opened 2021-03-25
Empty results no longer throws an error

### LINKED ISSUE #1101: Service Now API shows error if simply not found
If you search for a computer in the managed system popup and it is not found, you get an error message even though it is not an actual error.  Instead on not found just don't show the popup and have the text be something like "No computer matching your search could be found" instead of the no data to present generic thing.

image

wongband:
resolve #1101


srkirkland (approved):
good solution

Issue #1109, opened 2021-04-07
Search for an attribute in API


UCDFeedbackBot:
It would be handy to be able to search for equipment based on a certain attribute value over the API.
--------------------
Votes: 1
Author: rfoster
Voters: rfoster

PR #1110, opened 2021-04-07
Search Equipment Attributes

jSylvestre:
Issue #1109

### DIFF
--- a/Keas.Mvc/Controllers/Api/EquipmentController.cs
+++ b/Keas.Mvc/Controllers/Api/EquipmentController.cs
@@ -67,16 +67,10 @@ namespace Keas.Mvc.Controllers.Api
         [ProducesResponseType(typeof(IEnumerable<Equipment>), StatusCodes.Status200OK)]
         public async Task<IActionResult> SearchAttributes(string q)
         {
-            var attributes = from ea in _context.EquipmentAttributes
-                    .Where(a => a.Equipment.Team.Slug == Team && a.Equipment.Active &&
-                                (EF.Functions.Like(a.Value, q.EfStartsWith())))
-                    .AsNoTracking()
-                select ea.Equipment.Id;
-            var ids = await attributes.Distinct().ToArrayAsync();
-
             var equipment =
                 from eq in _context.Equipment
-                    .Where(x => x.Team.Slug == Team && x.Active && ids.Contains(x.Id))
+                    .Where(x => x.Team.Slug == Team && x.Active &&
+                                x.Attributes.Any(a => (EF.Functions.Like(a.Value, q.EfStartsWith()))))
                     .Include(x => x.Attributes)
                     .Include(x => x.Space).Include(x => x.Assignment)
                     .OrderBy(x => x.Assignment != null).ThenBy(x => x.Name)
--- a/Keas.Mvc/Controllers/Api/EquipmentController.cs
+++ b/Keas.Mvc/Controllers/Api/EquipmentController.cs
@@ -65,6 +65,38 @@ namespace Keas.Mvc.Controllers.Api
 
         [HttpGet]
         [ProducesResponseType(typeof(IEnumerable<Equipment>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> SearchAttributes(string q)
+        {
+            var attributes = from ea in _context.EquipmentAttributes
+                    .Where(a => a.Equipment.Team.Slug == Team && a.Equipment.Active &&
+                                (EF.Functions.Like(a.Value, q.EfStartsWith())))
+                    .Include(a => a.Equipment)
+                    .AsNoTracking()
+                select new
+                {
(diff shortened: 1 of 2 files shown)


Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -63,6 +63,38 @@ select new
...
+            var attributes = from ea in _context.EquipmentAttributes
+                    .Where(a => a.Equipment.Team.Slug == Team && a.Equipment.Active &&
+                                (EF.Functions.Like(a.Value, q.EfStartsWith())))
+                    .Include(a => a.Equipment)
+                    .AsNoTracking()
+                select new
+                {
+                    ea.Equipment.Id
+                };
+            var ids = await attributes.ToListAsync();
```
sprucely, line 79: You should be able to call `Distinct()` before `ToListAsync()` so that it is performed at the db instead of locally.

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -63,6 +63,38 @@ select new
...
         }
 
+        [HttpGet]
+        [ProducesResponseType(typeof(IEnumerable<Equipment>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> SearchAttributes(string q)
+        {
+            var attributes = from ea in _context.EquipmentAttributes
+                    .Where(a => a.Equipment.Team.Slug == Team && a.Equipment.Active &&
+                                (EF.Functions.Like(a.Value, q.EfStartsWith())))
+                    .Include(a => a.Equipment)
```
srkirkland, line 73: .Include doesn't do anything if you are going to do a `select new { id }` anyway

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -63,6 +63,36 @@ select new
...
+        {
+            var attributes = from ea in _context.EquipmentAttributes
+                    .Where(a => a.Equipment.Team.Slug == Team && a.Equipment.Active &&
+                                (EF.Functions.Like(a.Value, q.EfStartsWith())))
+                    .AsNoTracking()
+                select new
+                {
+                    ea.Equipment.Id
+                };
+            var ids = await attributes.Distinct().Select(a => a.Id).ToArrayAsync();
```
srkirkland, line 78: @jSylvestre Why are you doing a select here when you already projected in the previous `attributes` query?  Doesn't seem like it would do anything

Keas.Mvc/Controllers/Api/EquipmentController.cs
```diff
@@ -63,6 +63,36 @@ select new
...
+                from eq in _context.Equipment
+                    .Where(x => x.Team.Slug == Team && x.Active && ids.Contains(x.Id))
+                    .Include(x => x.Attributes)
+                    .Include(x => x.Space).Include(x => x.Assignment)
+                    .OrderBy(x => x.Assignment != null).ThenBy(x => x.Name)
+                    .AsNoTracking()
+                select new
+                {
+                    equipment = eq,
+                    label = eq.Id + ". " + eq.Name + " " + eq.SerialNumber,
```
srkirkland, line 90: is label something we use in other places? I don't remember ever seeing it before.  Is it more useful than just returning the equipmnet?

srkirkland (commented):
In general I think this could probably be done in a single query instead of one for the ids and then a second to get the equipment, but i'm not sure it'll be used enough to cause a real problem if that's hard to figure out

Keas.Mvc/Controllers/Api/EquipmentController.cs
jSylvestre: @srkirkland If I don't it gives me a compile error:

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1929	'<anonymous type: int Id>[]' does not contain a definition for 'Contains' and the best extension method overload 'MemoryExtensions.Contains<int>(ReadOnlySpan<int>, int)' requires a receiver of type 'ReadOnlySpan<int>'	Keas.Mvc	D:\Work\GitProjects\Peaks\Keas.Mvc\Controllers\Api\EquipmentController.cs	82	Active

Keas.Mvc/Controllers/Api/EquipmentController.cs
jSylvestre: I copied that from the search above:
https://github.com/ucdavis/Peaks/blob/master/Keas.Mvc/Controllers/Api/EquipmentController.cs#L47

Keas.Mvc/Controllers/Api/EquipmentController.cs
jSylvestre: Or I can just do the select correctly....

srkirkland (approved):

PR #1117, opened 2021-04-15
Default supervisor when person is created

### LINKED ISSUE #1100: Automatically Populate Supervisor Info
When adding a person to PEAKS if the supervisor field could be auto-populated that would be very helpful.

--------------------
Votes: 4
Author: ssmith
Voters: ssmith,cmderr,rcasteli,cheyenna

jSylvestre:
Supervisor must already be a person in the team
resolves #1100

### DIFF
--- a/Keas.Mvc/Controllers/Api/PeopleController.cs
+++ b/Keas.Mvc/Controllers/Api/PeopleController.cs
@@ -174,12 +174,11 @@ namespace Keas.Mvc.Controllers.Api
                 User = user,
                 Title = await _identityService.GetTitle(user.Iam)
             };
-            var supervisor = await _identityService.GetIamSupervisor(user.Iam);
-            if (supervisor != null)
+            var supervisorIam = await _identityService.GetIamSupervisor(user.Iam);
+            if (!string.IsNullOrWhiteSpace(supervisorIam))
             {
                 var superPerson = await _context.People
-                    .Where(x => x.Team.Slug == Team && x.UserId == supervisor.Id)
-                    .Include(x => x.Supervisor)
+                    .Where(x => x.Team.Slug == Team && x.User.Iam == supervisorIam)
                     .IgnoreQueryFilters()
                     .FirstOrDefaultAsync();
                 if (superPerson != null)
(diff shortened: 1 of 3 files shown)


Keas.Mvc/Controllers/Api/PeopleController.cs
```diff
@@ -174,6 +174,20 @@ public async Task<IActionResult> SearchUsers(string searchTerm)
...
+            if (supervisor != null)
+            {
+                var superPerson = await _context.People
+                    .Where(x => x.Team.Slug == Team && x.UserId == supervisor.Id)
+                    .Include(x => x.Supervisor)
+                    .IgnoreQueryFilters()
+                    .FirstOrDefaultAsync();
+                if (superPerson != null)
+                {
+                    person.SupervisorId = superPerson.Id;
```
srkirkland, line 187: I'm not sure it's a problem to set both but best practice is just to set the ID when doing these FK relationships

Keas.Mvc/Controllers/Api/PeopleController.cs
```diff
@@ -174,6 +174,20 @@ public async Task<IActionResult> SearchUsers(string searchTerm)
...
                 Title = await _identityService.GetTitle(user.Iam)
             };
+            var supervisor = await _identityService.GetIamSupervisor(user.Iam);
+            if (supervisor != null)
+            {
+                var superPerson = await _context.People
+                    .Where(x => x.Team.Slug == Team && x.UserId == supervisor.Id)
+                    .Include(x => x.Supervisor)
+                    .IgnoreQueryFilters()
+                    .FirstOrDefaultAsync();
```
srkirkland, line 184: Why the include? Why not just do the where and then project out `select()` just the Id that you need.  Much more efficient query for the same result.

srkirkland (approved):
Maybe i'm missing the edge case here but what if the `reportsToIAMID` person doesn't exist in our Users table?  We should create them right?  And then what if the user returned from the identity service isn't in our People table?

Keas.Mvc/Controllers/Api/PeopleController.cs
jSylvestre: That was a copy/paste mistake. Fixed

PR #1119, opened 2021-04-27
Get rid of those pesky leading spaces

### LINKED ISSUE #1118: Search for person to add them
A space in front of the email causes an error.
Need to trim the search value

jSylvestre:
Resolves #1118


srkirkland (changes requested):
Only truncate if the string is not null, otherwise it'll error out where before it would have worked

PR #1125, opened 2021-05-19
Add Team slug to Group Reports

jSylvestre:
EquipmentReport


sprucely (approved):
Haven't tested, but looks fine to me

PR #1146, opened 2022-01-05
Upload the loading gif

jSylvestre:


srkirkland (changes requested):
Nope we don't want those changes to the csproj.

jSylvestre:
Redoing...

PR #1147, opened 2022-01-06
Improve logging

sprucely:
Simplifies logging config, logs app startup, and adds more log context.

It's still logging to Stackify. Wasn't sure if I should remove that.


Keas.Jobs.Core/LogConfiguration.cs
```diff
@@ -37,58 +39,32 @@ public static void Setup(IConfigurationRoot configuration)
...
-                .WriteToStackifyCustom();
+                .WriteTo.Stackify();
 
-            return logConfig;
-        }
-
-        private static LoggerConfiguration WriteToStackifyCustom(this LoggerConfiguration logConfig)
-        {
-            if (!_loggingSetup)
+            if (esUrl?.StartsWith("http") ?? false)
```
srkirkland, line 57: for me this is confusing to read compared to `if (esUrl.HasValue && esUrl.StartsWith("http"))` because of the `?? false` part.  I wonder if there is an even easier way to write it with the new C# pattern matchers, something like `if (esUrl is { StartsWith("http") })`.  I just made that last one up though it's probably not even valid syntax but that's the idea.

sprucely: @srkirkland Doesn't look like they have method pattern matching yet; just property pattern matching. This is probably the closest we could get to what you're thinking...
```csharp
logConfig = esUrl switch
{
    var validUrl when validUrl.StartsWith("http") => logConfig.Enrich.WithProperty("Application", loggingSection.GetValue<string>("AppName"))
        .Enrich.WithProperty("AppEnvironment", loggingSection.GetValue<string>("Environment"))
        .WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri(validUrl))
        {
            IndexFormat = "aspnet-peaks-{0:yyyy.MM.dd}"
        }),
    _ => logConfig
};
```
It's a little better, but probably not worth it over just doing `if (esUrl.HasValue && esUrl.StartsWith("http"))`

srkirkland (approved):

PR #1152, opened 2022-01-12
Disable APM and remove redundant logging init code

sprucely:


Keas.Mvc/Startup.cs
```diff
@@ -218,11 +218,7 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory, IHostApplicationLifetime appLifetime)
         {
-            app.UseAllElasticApm(Configuration);
-            
-            // setup logging
-            app.ConfigureStackifyLogging(Configuration);
```
jSylvestre, line 224: Did you mean to delete these?

srkirkland: Agreed I would think you'd still need the serilog entry at least.

Keas.Mvc/Program.cs
```diff
@@ -89,7 +89,7 @@ static LoggerConfiguration CreateLogConfig(IConfigurationRoot configuration)
                 .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
                 // .MinimumLevel.Override("Microsoft.EntityFrameworkCore", LogEventLevel.Warning) // uncomment this to hide EF core general info logs
                 .MinimumLevel.Override("Microsoft.AspNetCore", LogEventLevel.Warning)
-                .Enrich.WithElasticApmCorrelationInfo()
+                //.Enrich.WithElasticApmCorrelationInfo()
```
srkirkland, line 92: Does it add much to the logs if we leave the correlation? I think if we remove our stackify logger we would want to keep this apm enriched.

srkirkland (changes requested):

Keas.Mvc/Startup.cs
sprucely: `ConfigureStackifyLogging()` is called in `Program.cs`
`loggerFactory.AddSerilog()` is the old way of doing it, and is handled internally by `IHostBuilder.UseSerilog()` in `Program.cs`

PR #1168, opened 2022-05-10
Once expiry date is more than 30 days old, the next notification date…

jSylvestre:
… will be 30 days away.

Instead of PR #1167


srkirkland (approved):
👍 .  Not sure this is important or interesting but if we wanted to keep the notifications coming on a regular cadence you could use `AddMonths` instead of `AddDays` so we'd always be sending the same day each month instead of an occasional off by one progressive thing.

PR #1169, opened 2022-05-10
Jcs/expiring report service

jSylvestre:


sprucely (approved):

Keas.Mvc/Services/ReportService.cs
```diff
@@ -448,5 +454,136 @@ public async Task<IList<PeopleLeavingWithAssetsModel>> PeopleLeavingWithAssets(G
...
+
+
+            var expiringAccess = await _context.AccessAssignments.Where(a => (showType == "All" || showType == "Access") &&
+                a.Access.Team.Groups.Any(g => g.GroupId == group.Id) && a.ExpiresAt <= expiresBefore)
+                .Include(a => a.Person).Include(a => a.Access).ThenInclude(a => a.Team).Select(ExpiringAccessProjection()).ToArrayAsync();
+
+            if (expiringAccess.Any())
+            {
+                model.ExpiringItems.AddRange(expiringAccess);
+            }
```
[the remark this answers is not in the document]
jSylvestre: I couldn't remember, so I checked just in case...

Keas.Mvc/Services/ReportService.cs
```diff
@@ -448,5 +454,136 @@ public async Task<IList<PeopleLeavingWithAssetsModel>> PeopleLeavingWithAssets(G
...
+
+            if (expiringKey.Any())
+            {
+                model.ExpiringItems.AddRange(expiringKey);
+            }
+
+            var expiringEquipment = await _context.Equipment.Where(a => (showType == "All" || showType == "Equipment") &&
+                a.Team.Groups.Any(g => g.GroupId == group.Id) &&
+                a.Assignment.ExpiresAt <= expiresBefore)
+                .Include(a => a.Team).Include(e => e.Assignment).ThenInclude(a => a.Person).Select(ExpiringEquipmentProjection()).ToArrayAsync();
```
srkirkland, line 518: What is the point of `.Include()` if you are just going to do a projection?  The projection selects out what it wants.  I'm curious to see the SQL here -- does it just ignore the include or does it grab of bunch of unnecessary "include" records and then throw them away?  Either way I think it can only hurt

Keas.Mvc/Services/ReportService.cs
```diff
@@ -448,5 +454,136 @@ public async Task<IList<PeopleLeavingWithAssetsModel>> PeopleLeavingWithAssets(G
...
+            return a => new ExpiringItemReportModel
+            {
+                Type = "Equipment",
+                ItemName = a.Name,
+                PersonName = a.Assignment.Person.Name,
+                ExpiresAt = a.Assignment.ExpiresAt.ToPacificTime(),
+                TeamSlug = a.Team.Slug,
+                DetailsLink = $"/{a.Team.Slug}/equipment/details/{a.Id}", //This is really only needed for the non group one?
+            };
+        }
```
srkirkland, line 574: I do like this model though, declaring an expression for a projection and then using it where needed instead of a one-time inliner

srkirkland (changes requested):

Keas.Mvc/Services/ReportService.cs
jSylvestre: Ah. good point. I should have realized that.

PR #1175, opened 2022-05-12
Non async GroupBy(

jSylvestre:


srkirkland (changes requested):
you should keep them all as ToArray().  so after the groupBy do `.ToArray()` or the async version.

PR #1178, opened 2022-05-16
Deal with missing/deactived person

jSylvestre:


jSylvestre:
For the non-group report that isn't using projection, the assigned to was just blank.

PR #1180, opened 2022-05-16
Check for assignments when delete person

### LINKED ISSUE #1179: Prevent API deactivate person if has assignments

jSylvestre:
Closes #1179

### DIFF
--- a/Keas.Mvc/Controllers/Api/PeopleAdminController.cs
+++ b/Keas.Mvc/Controllers/Api/PeopleAdminController.cs
@@ -99,30 +99,27 @@ namespace Keas.Mvc.Controllers.Api
 
             if (personActing.UserId == person.UserId)
             {
-                ModelState.AddModelError("User", "Don't delete yourself.");
-                return BadRequest(ModelState);
+                ModelState.AddModelError("User", "Don't delete yourself.");                
             }
-
-
-
             if (await _context.AccessAssignments.AnyAsync(a => a.PersonId == id))
             {
                 ModelState.AddModelError("AccessAssignments", "Remove Access Assignments first");
-                return BadRequest(ModelState);
             }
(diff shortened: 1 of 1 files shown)


srkirkland (approved):

Keas.Mvc/Controllers/Api/PeopleAdminController.cs
```diff
@@ -91,15 +91,41 @@ public async Task<IActionResult> Delete(int id)
...
+            if (await _context.EquipmentAssignments.AnyAsync(a => a.PersonId == id))
+            {
+                ModelState.AddModelError("EquipmentAssignments", "Remove Equipment Assignments first");
+                return BadRequest(ModelState);
+            }
+            if (await _context.WorkstationAssignments.AnyAsync(a => a.PersonId == id))
+            {
+                ModelState.AddModelError("WorkstationAssignments", "Remove Workstation Assignments first");
+                return BadRequest(ModelState);
+            }
```
sprucely, line 127: Maybe remove all the `return BadRequest(ModelState);` statements and instead increment a modelErrorCount, then return BadRequest if that error count > 0? That would give more info about what needs to be corrected. Otherwise it feels like a client developer would have to play api whack-a-mole.

sprucely (changes requested):

Keas.Mvc/Controllers/Api/PeopleAdminController.cs
sprucely: Or instead of using a modelErrorCount, just inspect ModelState for any errors.

jSylvestre: I figured it should be rare, and if the API consumer isn't checking this first then they will get the next error if there are multiple assignments.

sprucely (approved):

PR #1181, opened 2022-05-18
Jcs/job sql timeout

jSylvestre:


jSylvestre:
Nah.. fixed with indexing...

PR #1186, opened 2022-06-14
Jcs/inactive spaces with stuff

jSylvestre:


sprucely (approved):
Looks good. Definitely wouldn't want to attempt that query with LINQ. Unless I'm bored.

Issue #1195, opened 2022-06-27
DocuSign report of Completed forms


UCDFeedbackBot:
It would be helpful to have a report of completed documents. I sent out the same form for 42 Faculty. It's the only forms we have sent out so far. The only report available in PEAKS shows me 40 have not signed it. So it will take a lot of elimination to determine which two have signed it.
--------------------
Votes: 1
Author: jjvidaca
Voters: jjvidaca

PR #1196, opened 2022-06-29
Completed doc report

jSylvestre:
Issue #1195

### DIFF
--- a/Keas.Mvc/Models/ReportModels/CompletedDocsReportModel.cs
+++ b/Keas.Mvc/Models/ReportModels/CompletedDocsReportModel.cs
@@ -1,0 +1,21 @@
+using Keas.Core.Domain;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace Keas.Mvc.Models.ReportModels
+{
+    public class CompletedDocsReportModel
+    {
+        [DataType(DataType.Date)]
+        [DisplayFormat(DataFormatString = "{0:yyyy-MM-dd}", ApplyFormatInEditMode = true)]
+        [Display(Name = "Start Created Date")]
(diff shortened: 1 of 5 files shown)


Keas.Mvc/Models/ReportModels/CompletedDocsReportModel.cs
```diff
@@ -0,0 +1,21 @@
...
+        [DataType(DataType.Date)]
+        [DisplayFormat(DataFormatString = "{0:yyyy-MM-dd}", ApplyFormatInEditMode = true)]
+        [Display(Name = "Start Created Date")]
+        public DateTime Start {get;set; }
+
+        [DataType(DataType.Date)]
+        [DisplayFormat(DataFormatString = "{0:yyyy-MM-dd}", ApplyFormatInEditMode = true)]
+        [Display(Name = "End Created Date")]
+        public DateTime End { get; set; }
+        public List<Document> docs {get;set;}
```
srkirkland, line 19: capital `D` (PascalCase for props)

Keas.Mvc/Views/Report/CompletedDocuments.cshtml
```diff
@@ -0,0 +1,105 @@
...
+                @foreach (var item in Model.docs)
+                {
+                    <tr>
+                        <td>@Html.DisplayFor(a=> item.Person.Name)</td>
+                        <td>@Html.DisplayFor(a=> item.Name)</td>
+                        <td>
+                            @item.CreatedAt.ToPacificTime().ToString("d")
+                        </td>     
+                        <td>
+                            @item.CompletedAt.Value.ToPacificTime().ToString("d")
```
srkirkland, line 68: should check that HasValue == true before just using value, even if we know it should never be null

PR #1252, opened 2023-03-30
Upgrade to net6 and update package versions

sprucely:


Keas.Mvc/Startup.cs
```diff
@@ -227,7 +225,6 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerF
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
-                app.UseDatabaseErrorPage();
```
jSylvestre, line 230: Deleted this on purpose?

jcubbage: Might have gotten warning on that line. It is obsolete and will be removed. Looks like there is a preferred method: https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.databaseerrorpageextensions.usedatabaseerrorpage?view=aspnetcore-6.0

sprucely: Yeah, there's a new `services.AddDatabaseDeveloperPageExceptionFilter();` that adds a db exception handler to the developer exception page stuff.

jSylvestre (approved):

srkirkland (approved):
🚢

PR #1258, opened 2023-05-30
Set minimum node version

sprucely:
The node version badge won't display correctly until the PR is merged to master.


jSylvestre (approved):
If this is safe, then sure `engine-strict=true`

srkirkland (approved):

PR #1267, opened 2023-06-27
Rh/name change

### LINKED ISSUE #1202: Finding people after a name change
We had a user change their name & email address after some life event. We couldn't find them afterwards (because we didn't know about the name change), but got a warning that we couldn't add the person as a new entry (because duplicate kerb ID). When that happens, system should provide link to the person in the team.
--------------------
Votes: 1
Author: jscub
Voters: jscub

laholstege:
closes #1202 
adds link when user already exists, routes to person details on click
the button to do so is a little off from the info string stating the user is already active. maybe a @cydoval problem? :) 

image

### DIFF
--- a/Keas.Mvc/ClientApp/src/components/People/CreatePerson.tsx
+++ b/Keas.Mvc/ClientApp/src/components/People/CreatePerson.tsx
@@ -16,14 +16,6 @@ interface IProps {
   closeModal: () => void;
 }
 
-interface IState {
-  error: IValidationError;
-  moreInfoString: string; // for explaining results, e.g. if person is new or inactive
-  person: IPerson;
-  submitting: boolean;
-  validState: boolean;
-}
-
 const CreatePerson = (props: IProps) => {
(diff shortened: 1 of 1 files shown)


Keas.Mvc/ClientApp/src/components/People/CreatePerson.tsx
```diff
@@ -33,6 +28,9 @@ const CreatePerson = (props: IProps) => {
   const [person, setPerson] = useState<IPerson>(null);
   const [submitting, setSubmitting] = useState<boolean>(false);
   const [validState, setValidState] = useState<boolean>(false);
+  const [exisitngPerson, setExisitingPerson] = useState<IPerson>(null);
```
srkirkland, line 31: spell check ❌

srkirkland (changes requested):

jSylvestre (approved):
I just say that is how they spell it in Canada :)

srkirkland (approved):

PR #1270, opened 2023-07-03
add make, model, spaces filters to equipment table

### LINKED ISSUE #1137: View equipment using additional fields
When viewing my department's equipment, it would be nice to view additional fields to be able to filter or sort by those fields like Make, Model, Room, etc. 
--------------------
Votes: 1
Author: viendih
Voters: viendih

laholstege:
closes #1137 

will redo some of this when i make one big fancy search box. cleaned up code to use shared components instead of having multiple components that were just a custom typeahead. now just either a typeahead with custom options (e.g. attributes, make, model, etc) or one with defined options that we pull from the server (tags, types, etc)

image


srkirkland (approved):
looks great, much cleaner using the shared search components 🚢

Issue #1249, opened 2023-03-16
Search strings in Attributes


UCDFeedbackBot:
Just FYI when searching in attributes field.
The string has to be exact in order to show correct results.

--------------------
Votes: 2
Author: jlutang
Voters: vanslyke,jlutang

PR #1275, opened 2023-07-06
fix broken attribute filtering

laholstege:
forgot to remove .label, filter is always an array of strings now (oops). also this was confusing me when i was debugging, so i tried to clean it up. accepts partial matches now, e.g. 'ma' matches 'mac'


jSylvestre (approved):
Didn't test, but looks ok.
Let me know if you want me to run it.

srkirkland (approved):

laholstege:
closes #1249

PR #1282, opened 2023-07-19
Fix reverse apostrophe issue and empty sections in expiring items email

sprucely:
FixXML converts character entity names (e.g. `&copy;`) to their corresponding character entity number (e.g. `&#169;`) so that it will parse successfully as valid xml. I don't think we use any of those, so it's not needed. And it was converting the reverse apostrophe `’` to `&#x2019;`. I think that `x` in there makes it invalid as a character entity number


jSylvestre (approved):
Didn't test, but looks right.

PR #1283, opened 2023-07-26
Cyd/cyd issues

cydoval:
addressing some outstanding issues and homepage redesign


sprucely (approved):
Tested. Looks good to me.

PR #1289, opened 2023-08-09
Rh/add assignment from unassigned asset

### LINKED ISSUE #1265: Add assignment from unassigned asset
looking at asset, ex. equipment.  you can update an existing assignment but not add a new one without exiting.

laholstege:
closes #1265 

always shows assignment details, links to assign modal so you can update without closing 

image


jSylvestre (approved):
Only tested Equipment.
Seems ok to me.

Issue #1273, opened 2023-07-03
can't unselect attribute/tags when creating/editing without selecting an option


laholstege:
![image](https://github.com/ucdavis/Peaks/assets/27025973/488b10d2-da3f-473a-9e22-456218efae4f)

laholstege:
will be fixed when we update react-bootstrap-typeahead to at least v5.2.2
https://github.com/ericgio/react-bootstrap-typeahead/issues/695

laholstege:
bug only happens inside of a modal--might be a reactstrap or bootstrap related issue?

laholstege:
https://github.com/ericgio/react-bootstrap-typeahead/issues/703

PR #1291, opened 2023-08-16
Rh/update typeahead

laholstege:
update react-bootstrap-typeahead and generalize async typeahead so it is shared like the other typeaheads

having a weird problem with the stylings after updating: 
image
@jSylvestre can you verify this happens for you too?


laholstege:
does not fix #1273

jSylvestre (approved):

PR #1292, opened 2023-08-16
unsupervise on person delete

### LINKED ISSUE #1089: Deleted supervisor
We don't have any checks preventing a supervisor from being deleted.
It doesn't really break anything having this like this.

laholstege:
closes #1089


jSylvestre (approved):
Didn't test, but looks solid

PR #1309, opened 2023-09-13
Jcs/search iam person

jSylvestre:
Ok, this works. If we want to use it we would need a job that runs after hours nightly.
We would probably also need something to try and update all existing users, keeping in mind that users may no longer be active in IAM.

Or... we just start running this and don't worry about old data.


laholstege (approved):

sprucely (approved):
Didn't test, but looks fine.

sprucely (approved):

PR #1313, opened 2023-09-21
Jcs/slim service

jSylvestre:


sprucely (changes requested):
Didn't test, but looks good. Yeah, just make a copy of the `SendMail` job in the pipeline yml andtweak it to point to the new job.

PR #1323, opened 2023-09-29
Fix check for logging

jSylvestre:


jSylvestre:
Actually, this is what I wanted to check...  
Ones missing on test also had no contact details (kerbs too)

PR #1327, opened 2023-10-04
Rh/make access consistent

laholstege:
sorry this one is so long! there was a lot to do to get it in line with the rest of our code. lots of state stuff in react–i don't expect that will be easy to follow reading but i tried to make it as clear as i could. if you could test and see if you find bugs on the /access/list as well as /person/details that would be great 

makes access components match the pattern of other assets. 
- adds details modal for individual access assignments, in same pattern as key serials
- makes 'em pretty

- removes nested/recursive access state–at most, `access.assignments[0].access` will be populated, instead of `access.assignments[0].access.assignments[]...` 
- when on /access/ (`AccessContainer`) we will have `access` with fully populated `.assignments` (which will not have a nested access)
- when on /people/ (`AccessAssignmentContainer`) we will have `assignments[]` with a populated `.access` (which will not have nested assignments)
-when on /people/ (`AccessAssignmentContainer`, calling `access/listAssigned`, we only pull the assignments for that person, and pass `numberOfAssignments` for the list to use
- fixes a lot of small bugs, bad handling of state, clarifies names

known weird/less than ideal behavior
- when on /person/details, if you assign an access it has an incorrect "number of assignments" value in the table until refresh. ~~this is because in AccessController, we do not pass back the accesses, so `assignments.length` on the frontend is wrong. could pass back all the assignments (seems unnecessary), could add a property of `numberOfAssignments`, could just delete the column in the list, or just do nothing about it.~~
~~- still have some nested state under person details, but intending to fix this next~~


Keas.Mvc/ClientApp/src/components/Equipment/EquipmentEditValues.tsx
```diff
@@ -80,17 +80,17 @@ const EquipmentEditValues = (props: IProps) => {
 
   return (
     <div>
-      {props.disableEditing && props.openEditModal && (
-        <div className='row justify-content-between'>
-          <h3>Equipment Details</h3>
+      <div className='row justify-content-between'>
+        <h3>Equipment Details</h3>
```
jSylvestre, line 84: I don't know that this is needed if we are creating a new equipment.
It has the Create New Equipment title when we do that.
image

jSylvestre (approved):

PR #1337, opened 2023-10-07
Rh/use effect dependency arrays

### LINKED ISSUE #1326: react useEffect dependency arrays warning

laholstege:
closes #1326 
also solves confirm warnings
and adds confirm for revoking key serial


jSylvestre:
@laholstege Was this supposed to fix it so a toast error shows up if the equipment detail id couldn't be found?

laholstege:
@jSylvestre no, this is just the ts warnings. all i added was an alert "Are you sure you want to revoke?" to key serials

PR #1350, opened 2023-11-20
fix useEffect() infinite loop

laholstege:
eslint wants the functions we're using in our useEffect() to be part of the dependency array, but our functions are `() => {}` so it changes every time and loops. stupid to use the entire props in the dep array regardless, that's my bad. i believe this is a case where we'd want to use `useCallback()` but i'll get that working in a different PR


srkirkland (approved):
looks good, makes sense using the whole props obj as dep can cause issues.  funny we don't run into this more often.  I agree useCallback is probably the answer but that would probably require a lot of housekeeping to remember to do it in the parent.

jSylvestre (approved):

PR #1351, opened 2024-03-15
manually set KeySerialAssignmentId to null on revoke

laholstege:


jSylvestre (approved):
Didn't test, but looks good.

srkirkland (approved):

PR #1364, opened 2024-10-04
Don't try to call into KFS to validate KFS org

jSylvestre:


sprucely (approved):
Looks fine as a temporary fix.

PR #1368, opened 2025-02-18
Try to prevent dup people

jSylvestre:


Keas.Mvc/Services/IdentityService.cs
```diff
@@ -402,13 +412,18 @@ public async Task<string> BulkLoadPeople(string ppsCode, string teamslug, string
                         };
                         _context.People.Add(newPerson);
                         await _context.SaveChangesAsync(); //Need to save person so it has an id for notification below
-                        await _notificationService.PersonUpdated(newPerson, team, teamslug, actorName, actorId, PersonNotification.Actions.Added, $"Bulk Load from PPS code: {ppsCode}");
+                        await _notificationService.PersonUpdated(newPerson, team, teamslug, actorName, actorId, PersonNotification.Actions.Added, $"Bulk Load from PPS code: {ppsCode}");
+                        somethingUpdated = true;
```
sprucely, line 416: Maybe don't set `somethingUpdated = true` here, since it is explicitly saving changes? Or does `_notificationService.PersonUpdated` make an additional change that also needs to be saved?

srkirkland (approved):

sprucely (changes requested):

sprucely (approved):
NVM

PR #1378, opened 2025-11-17
Updates FIS Org to use OrgCode only

jSylvestre:
Updates the financial organization to only require the OrgCode instead of Chart and OrgCode.
Removes the Chart field from the model and database. Chart is defaulted to "X" in the backend.
Increases OrgCode length to 6.

Relates to JCS/FisOrgCode


jSylvestre:
Prod database has been updated.

jSylvestre:
Don't merge this until we have to.
This has to be coordinated with a change to the ETL process.
For that, we would want to populate an X in the chart and the full 6 character AE org code

